PR #19, opened 2022-08-02
Add initial structure

jacobtomlinson:
Playing around a bit with structure here. This is mostly just thinking out loud, feedback very welcome.

I've mashed together some of the different page types we have currently into a rough taxonomy. However as I see it there are three different types of page here:

- Tool specific (`dask-kubernetes`, `dask-helm`, `rapids conda`, ...)
- Platform specific (Sagemaker, Databricks, KubeFlow, HPC + SLURM, HPC + PBS, ...)
- Usage Scenario (ML platform, Single VM, Multi-user, ...)

These types of page are messy and overlapping. For example KubeFlow and Kubernetes + Dask Operator have a lot of overlap.


jacobtomlinson:
Perhaps it makes sense to split the docs into two sections:

- Tools
- Guides

Tools would act as how-to and reference for each piece of deployment tooling. For example how to install the Dask Helm Chart with RAPIDS configuration, or how to use `dask-cloudprovider` to launch an `EC2Cluster` with RAPIDS images. This would be an exhaustive list of every tool that can deploy rapids.

Guides would then be more opinionated and possibly feel like a choose your own adventure linking out to specific tool pages. For example we could have guides on `Sagemaker`, `KubeFlow`, `HPC + SLURM`, etc which would cover explanation and context, any platform specific steps and links to tools that can be used in that workflow.

Given there are multiple ways to deploy things I'm tempted to have flowcharts for each platform guide that helps steer folks to the right tool for them. For example `HPC + SLURM` could have an example job submit script and also link to tool pages for `dask-jobqueue` and `dask-mpi` (and maybe even `dask-ssh`).

jacobtomlinson:
Ok I've split this out into two sections, platforms and tools. This way we can keep reusable information relating to specific tools in one place. 

I still want to capture usage scenarios somehow, I wonder if we can add tags or something.

Issue #10, opened 2022-06-29
Kubeflow + RAPIDS


mmccarty:

jacobtomlinson:
This blog post should be a useful foundation for this

https://jacobtomlinson.dev/posts/2022/using-dask-on-kubeflow-with-the-dask-kubernetes-operator/

jacobtomlinson:
This post will also be useful for running KubeFlow on a workstation with kind

https://jacobtomlinson.dev/posts/2022/running-kubeflow-inside-kind-with-gpu-support/

jacobtomlinson:
https://github.com/rapidsai/deployment/pull/21

Issue #8, opened 2022-06-29
Dask Gateway + RAPIDS


mmccarty:

mroeschke:
Should this be shown using dask gateway on Kubernetes with RAPIDS?

mmccarty:
Yes

Issue #7, opened 2022-06-29
Dask Helm Charts + RAPIDS


mmccarty:
Write a documentation page on installing the Dask Helm Chart but with GPUs and RAPIDS.

Preparation:

- Get a Kubernetes cluster (use kind or GKE)
- Install helm
- Try installing the Dask helm chart
  - https://github.com/dask/helm-chart/tree/main/dask
- Check you can get into Jupyter and use the Dask cluster

Documentation outline:

- Get Kubernetes cluster
- Install the Dask helm repo
- Create a config file that sets RAPIDS things
  - Docker image
  - Worker startup command (`dask-cuda-worker`
  - Resources should include a GPU
- Check you can get into Jupyter and use the Dask cluster
- Run through the 10 mins to dask-cudf to show it works

isVoid:
Would love to have a bit more information on step 3: `Create a config file that sets RAPIDS things
`. 

My understanding of `dask-helms-chart` is that it would setup all the scheduler and worker pod with the dask images. There are a couple things we need to do:
1. Replace the dask images with rapids run time images.
2. Add `DISABLE_JUPYTER="true" to worker and scheduler specs.
3. Explicitly request GPU for the scheduler and worker pods.
4. Replace the startup commands for worker pods with `dask-cuda-worker`.

I have a set of working helm-chart derived from dask-helms-chart hosted here: https://github.com/isVoid/helm-chart/tree/rapids/rapids but it looks very much like rapids-helms-chart that just got deprecated. Is there a better way to apply the above changes without having to host a separate set of helm charts?

jacobtomlinson:
Your list is pretty much spot on, we should create an example config file that sets the things you listed. We should verify it works and then users can just copy and paste. The [valyes.yaml](https://github.com/rapidsai/helm-chart/blob/main/rapidsai/values.yaml) file from the deprecated RAPIDS chart is a good place to go for inspiration, but it hasn't been updated since RAPIDS `0.17` so things may not work quite the same now.

Instead of hosting a separate chart we just need [this documentation page](https://github.com/rapidsai/deployment/blob/main/source/tools/kubernetes/dask-helm-chart.md) to be populated with instructions on how to use the Dask Helm Chart with RAPIDS. If you create a config that works and then put it on that page with some explanation around what each option is doing and why we need to do it, and some instructions on how to do it, that would be awesome!

PR #22, opened 2022-08-11
Add general Kubernetes page

jacobtomlinson:
I've added general info to the Kubernetes page.

My goal for this page is to be a general landing point for folks looking for Kubernetes info, but I expect most sections will be very short and mostly help folks choose which tool they want for their use case and then just link off to each `/tools/foo` page for more detailed information.

I've also started the page with a small example of using the RAPIDS container by creating a single `Pod` and `Service`. This is basically a k8s version of doing `docker run ...` and getting a single RAPIDS container. (thanks @jarmak-nv for the example)


jacobtomlinson:
xref #5 #6 #7 #8 #10

jacobtomlinson:
Thanks for the review @mroeschke, merging.

PR #23, opened 2022-08-11
Add Dask operator docs

### LINKED ISSUE #6: Dask Operator + RAPIDS

jacobtomlinson:
Added a page on configuring RAPIDS Dask clusters with the [Dask Operator](https://kubernetes.dask.org/en/latest/operator.html).

I also added intersphinx config to I could easily crosslink to the Dask docs and I added [sphinxcontrib-mermaid](https://pypi.org/project/sphinxcontrib-mermaid/) for pretty diagrams.

Closes #6


source/tools/kubernetes/dask-operator.md
```diff
@@ -1 +1,381 @@
...
+          port: 8786
+          targetPort: "tcp-comm"
+        - name: http-dashboard
+          protocol: TCP
+          port: 8787
+          targetPort: "http-dashboard"
+```
+
+This example shows using a `ClusterIP` service which will not expose the Dask cluster outside of Kubernetes. If you prefer you could set this to
+`LoadBalancer` or `NodePort` to make this externally accessible.
```
mroeschke, line 280:
```suggestion
[`LoadBalancer`](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer) or [`NodePort`](https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport) to make this externally accessible.
```

Since you have the reference below too

source/tools/kubernetes/dask-operator.md
```diff
@@ -1 +1,381 @@
...
+
+## Example using `KubeCluster`
+
+In additon to creating clusters via `kubectl` you can also do so from Python with {class}`dask_kubernetes.experimental.KubeCluster`. This class implements the Dask Cluster Manager interface and under the hood creates and manages the `DaskCluster` resource for you.
+
+```python
+from dask_kubernetes.experimental import KubeCluster
+
+cluster = KubeCluster(name="rapids-dask",
+                      image="rapidsai/rapidsai-core:22.06-cuda11.5-runtime-ubuntu20.04-py3.9",
```
mroeschke, line 331: Might be good to explicitly state `n_workers=3` (or 2 if you want to match the `kubectl` example above). It wasn't immediately clear to me why this created 3 workers and the `kubectl` example had 2.

jacobtomlinson:
Thanks for the review @mroeschke!

mroeschke (approved):

PR #33, opened 2022-08-18
Create docs build.sh

AyodeAwe:
This creates the required build script for the deployment of the docs generated in this repository.

### DIFF
--- a/ci/docs/build.sh
+++ b/ci/docs/build.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+# Copyright (c) 2020-2022, NVIDIA CORPORATION.
+#################################
+# RAPIDS Deployment Docs build script for CI #
+#################################
+
+if [ -z "$PROJECT_WORKSPACE" ]; then
+    echo ">>>> ERROR: Could not detect PROJECT_WORKSPACE in environment"
+    echo ">>>> WARNING: This script contains git commands meant for automated building, do not run locally"
+    exit 1
+fi
+
+export DOCS_WORKSPACE="$WORKSPACE"
+export PROJECTS=(deployment)
+
+make html
+
+#Commit to Website
+cd $DOCS_WORKSPACE
+
+for PROJECT in ${PROJECTS[@]}; do
+    if [ ! -d "api/$PROJECT/$BRANCH_VERSION" ]; then
+        mkdir -p api/$PROJECT/$BRANCH_VERSION
+    fi
+    rm -rf $DOCS_WORKSPACE/api/$PROJECT/$BRANCH_VERSION/*
+done
+
+
+mv $PROJECT_WORKSPACE/build/html/* $DOCS_WORKSPACE/api/deployment/$BRANCH_VERSION
--- a/ci/docs/build.sh
+++ b/ci/docs/build.sh
@@ -10,8 +10,12 @@ if [ -z "$PROJECT_WORKSPACE" ]; then
     exit 1
 fi
 
-export DOCS_WORKSPACE="$WORKSPACE"
+export DOCS_WORKSPACE="$WORKSPACE/docs"
 export PROJECTS=(deployment)
+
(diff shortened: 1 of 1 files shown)


ci/docs/build.sh
```diff
@@ -0,0 +1,29 @@
...
+
+for PROJECT in ${PROJECTS[@]}; do
+    if [ ! -d "api/$PROJECT/$BRANCH_VERSION" ]; then
+        mkdir -p api/$PROJECT/$BRANCH_VERSION
+    fi
+    rm -rf $DOCS_WORKSPACE/api/$PROJECT/$BRANCH_VERSION/*
+done
+
+
+mv $PROJECT_WORKSPACE/build/html/* $DOCS_WORKSPACE/api/deployment/$BRANCH_VERSION
```
AyodeAwe, line 29: I'm not 100% sure the deployment docs should go in the [api section](https://docs.rapids.ai/api) of the website. I just followed with the same location of the other libraries.

Thoughts?

ci/docs/build.sh
```diff
@@ -0,0 +1,29 @@
...
+
+make html
+
+#Commit to Website
+cd $DOCS_WORKSPACE
+
+for PROJECT in ${PROJECTS[@]}; do
+    if [ ! -d "api/$PROJECT/$BRANCH_VERSION" ]; then
+        mkdir -p api/$PROJECT/$BRANCH_VERSION
+    fi
```
ajschmidt8, line 24:
```suggestion
    mkdir -p api/$PROJECT/$BRANCH_VERSION
```

I know this was likely copy/pasted from another repo but we can simplify this logic here. `mkdir -p` will still exit with `0` if a folder already exists, so we don't need to guard it in an `if` statement.

ci/docs/build.sh
ajschmidt8: Based on Jacob's input, I think it should be fine to include this in the `/api` section of the site.

ci/docs/build.sh
```diff
@@ -0,0 +1,29 @@
...
+# RAPIDS Deployment Docs build script for CI #
+#################################
+
+if [ -z "$PROJECT_WORKSPACE" ]; then
+    echo ">>>> ERROR: Could not detect PROJECT_WORKSPACE in environment"
+    echo ">>>> WARNING: This script contains git commands meant for automated building, do not run locally"
+    exit 1
+fi
+
+export DOCS_WORKSPACE="$WORKSPACE"
```
ajschmidt8, line 13:
```suggestion
export DOCS_WORKSPACE="$WORKSPACE/docs"
```

this should be consistent with the `cudf` path below

- https://github.com/rapidsai/cudf/blob/288c81fd59d1b28541acb39235b314476f8a846a/ci/docs/build.sh#L13

ajschmidt8 (commented):
Left a few minor comments. Also, I think we need to include the following statements to ensure that the `conda` environment is activated before we run `make html`. The `conda` environment will include all of the dependencies that are necessary to build the docs.

```sh
export PATH=/conda/bin:/usr/local/cuda/bin:$PATH
. /opt/conda/etc/profile.d/conda.sh
conda activate rapids
```

ci/docs/build.sh
AyodeAwe: I had initially had it that way but removed it as I realized that there is no `docs` directory in the root directory of this repository. Am I missing something?

ajschmidt8: Yes, the Jenkins job below clones the https://github.com/rapidsai/docs/ repo into the `$WORKSPACE/docs` directory.

- https://6abc.com/northern-lights-forecast-philadelphia-pennsylvania-nj/12137915/

AyodeAwe: Oops, thanks.

ajschmidt8 (approved):

PR #45, opened 2022-10-18
Add build and deploy workflow

### LINKED ISSUE #46: Automate releasing docs
xref https://github.com/rapidsai/ops/issues/2304

jacobtomlinson:
Add workflow to build docs and optionally deploy it.

All PRs will have the docs built as part of general testing that things are working ok.

If on `main` in the upstream `rapidsai/deployment` repo built HTML files are also committed to [the docs repo](https://github.com/rapidsai/docs/tree/gh-pages/deployment/) for publishing. Regular commits to `main` will be put at the `/deployment/nightly` path and tagged commits to the `/deployment/stable` path.

Closes #46

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -26,7 +26,7 @@ jobs:
           miniforge-variant: Mambaforge
           use-mamba: true
           activate-environment: deployment-docs-dev
-          environment-file: conda/environments/deployment_docs_cuda11.5.yml
+          environment-file: conda/environments/deployment_docs.yml
 
       - name: Build
         run: make html
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -1,8 +1,9 @@
 name: Build and deploy
 on:
   push:
-    branches:
-      - main
+    # TODO: Uncomment before merging
+    # branches:
+    #   - main
   pull_request:
 
 # Required shell entrypoint to have properly activated conda environments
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,45 @@
+name: Build and deploy
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
ajschmidt8, line 6: based on the `refs/tags` reference at the bottom of this file, should these events include `tags`?

e.g. https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs

jacobtomlinson: The workflow needs to run on all push events to the main branch, so tags and regular commits. So I left it off there and then check later to decide what path to push the artefacts to.

Does that sound right?

ajschmidt8: Hmm. I'm not sure I understand. It looks to me as if this workflow would _never_ run on any `tag` pushes since it's missing something like:

```
on:
  push:
    tags:        
      - "*"
```

jacobtomlinson: Ahhh I get it, you mean in addition to the `main` branch. I thought you meant instead of.

jacobtomlinson: I've added this now.

ajschmidt8 (approved):

PR #59, opened 2022-10-31
Remove CUDA env

### LINKED ISSUE #18: Environment creation failed on non-GPU system
I tried building these docs locally on my MacBook but got the following error when creating the environment.

```console
$ conda env create -f conda/environments/deployment_docs_cuda11.5.yml
Collecting package metadata (repodata.json): done
Solving environment: failed

ResolvePackageNotFound:
  - rapids=22.02
  - cudatoolkit=11.5
```

This makes sense as the system doesn't have a GPU, but I'm curious whether we will actually need RAPIDS and a GPU capable system to build these docs? If we do then we should add a note to the README.

jacobtomlinson:
Locally and in CI I'm using the `conda/environments/deployment_docs.yml` environment. We aren't using the CUDA one for anything right now and it is currently pinned to RAPIDS 22.02 and CUDA 11.5 and that will only get more stale.

@mroeschke suggested in #18 that it may be useful in the future if we want to auto document some RAPIDS things, but for the time being I think it is adding confusion and slowly going stale due to lack of use.

I propose we make the non-CUDA one the default in the README and delete the CUDA one for now and reinstate it in the future if it becomes valuable again.

Closes #18


mroeschke (approved):

jacobtomlinson:
Thanks for the review @mroeschke

PR #63, opened 2022-11-04
Add page on Multi-Instance GPU (MIG)

pentschev:
Recently the interest for deploying Dask with Multi-Instance GPUs (MIG) has grown, therefore we add a getting started page, also communicating some of the limitations of MIG in an HPC context.

### DIFF
--- a/source/index.md
+++ b/source/index.md
@@ -46,6 +46,17 @@ Use RAPIDS on high performance computers and supercomputers.
 ````
 
 ````{grid-item-card}
+:link: mig
+:link-type: doc
+{fas}`border-all;sd-text-primary` Multi-Instance GPUs
+^^^
+Use RAPIDS with Multi-Instance GPUs
+
+{bdg-primary}`Dask Cluster`
+{bdg-primary}`XGBoost with Dask Cluster`
+````
+
+````{grid-item-card}
 :link: platforms/index
 :link-type: doc
 {fas}`network-wired;sd-text-primary` Platforms
@@ -92,6 +103,7 @@ See our [example notebooks repo](https://github.com/rapidsai/cloud-ml-examples) 
 local
 cloud/index
 hpc
+mig
 platforms/index
 ```
 
--- a/source/mig.md
+++ b/source/mig.md
@@ -1,8 +1,8 @@
 # Multi-Instance GPU (MIG)
 
-Multi-Instance GPU is a technology that allows partitioning a single GPU into multiple instances, making each one seem as a completely independent GPU. Each instance then receives a certain slice of the GPU computational resources and a pre-defined block of memory that is detached from the other instances by on-chip protections.
+[Multi-Instance GPU](https://www.nvidia.com/en-us/technologies/multi-instance-gpu/) is a technology that allows partitioning a single GPU into multiple instances, making each one seem as a completely independent GPU. Each instance then receives a certain slice of the GPU computational resources and a pre-defined block of memory that is detached from the other instances by on-chip protections.
 
-Due to the protection layer to make MIG secure, certain limitations exist. One such limitation that is generally important for HPC applications is the lack of support for CUDA Inter-Process Communication (IPC), which enables transfers over NVLink and NVSwitch to greatly speed up communication between physical GPUs. When using MIG, NVLink and NVSwitch are thus completely unavailable, forcing the application to take a more expensive communication channel via the system (CPU) memory.
+Due to the protection layer to make MIG secure, certain limitations exist. One such limitation that is generally important for HPC applications is the lack of support for [CUDA Inter-Process Communication (IPC)](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#interprocess-communication), which enables transfers over NVLink and NVSwitch to greatly speed up communication between physical GPUs. When using MIG, [NVLink and NVSwitch](https://www.nvidia.com/en-us/data-center/nvlink/) are thus completely unavailable, forcing the application to take a more expensive communication channel via the system (CPU) memory.
 
 Given limitations in communication capability, we advise users to first understand the tradeoffs that have to be made when attempting to setup a cluster of MIG instances. While the partitioning could be beneficial to certain applications that need only a certain amount of compute capability, communication bottlenecks may be a problem and thus need to be thought of carefully.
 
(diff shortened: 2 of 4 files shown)


source/mig.md
```diff
@@ -0,0 +1,76 @@
+# Multi-Instance GPU (MIG)
+
+Multi-Instance GPU is a technology that allows partitioning a single GPU into multiple instances, making each one seem as a completely independent GPU. Each instance then receives a certain slice of the GPU computational resources and a pre-defined block of memory that is detached from the other instances by on-chip protections.
```
jacobtomlinson, line 3: Is there a MIG product page we could like "Multi-Instance GPU" to?

source/mig.md
```diff
@@ -0,0 +1,76 @@
...
+
+Due to the protection layer to make MIG secure, certain limitations exist. One such limitation that is generally important for HPC applications is the lack of support for CUDA Inter-Process Communication (IPC), which enables transfers over NVLink and NVSwitch to greatly speed up communication between physical GPUs. When using MIG, NVLink and NVSwitch are thus completely unavailable, forcing the application to take a more expensive communication channel via the system (CPU) memory.
```
jacobtomlinson, line 5: Could we link "CUDA Inter-Process Communication (IPC)" somewhere useful? And maybe "NVLink and NVSwitch"?

source/mig.md
```diff
@@ -0,0 +1,76 @@
...
+
+Given limitations in communication capability, we advise users to first understand the tradeoffs that have to be made when attempting to setup a cluster of MIG instances. While the partitioning could be beneficial to certain applications that need only a certain amount of compute capability, communication bottlenecks may be a problem and thus need to be thought of carefully.
+
+## Dask Cluster
+
+Dask clusters of MIG instances are supported via Dask-CUDA as long as all MIG instances are identical with respect to memory. Much like a cluster of physical GPUs, mixing GPUs with different memory sizes is generally not a good idea as Dask may not be able to balance work correctly and eventually could lead to more frequent out-of-memory errors.
```
jacobtomlinson, line 11: Could we link "Dask-CUDA" to the docs?

source/mig.md
```diff
@@ -0,0 +1,76 @@
...
+
+For example, partitioning two GPUs into 7 x 10GB instances each and setting up a cluster with all 14 instances should be ok. However, partitioning one of the GPUs into 7 x 10GB instances and another with 3 x 20GB should be avoided.
+
+Unlike for a system composed of unpartitioned GPUs, Dask-CUDA cannot automatically infer the GPUs to be utilized for the cluster. In a MIG setup, the user is then required to specify the GPU instances to be used by the cluster. This is achieved by specifying either the `CUDA_VISIBLE_DEVICES` environment variable for either `LocalCUDACluster` or `dask-cuda-worker`, or the homonymous argument for `LocalCUDACluster`.
```
jacobtomlinson, line 15: Maybe we should use intersphinx to link `LocalCUDACluster` to the reference in the `dask-cuda` docs?

source/mig.md
```diff
@@ -0,0 +1,76 @@
...
+
+```bash
+CUDA_VISIBLE_DEVICES="MIG-41b3359c-e721-56e5-8009-12e5797ed514,MIG-65b79fff-6d3c-5490-a288-b31ec705f310,MIG-c6e2bae8-46d4-5a7e-9a68-c6cf1f680ba0" dask-cuda-worker scheduler.json # --other-arguments
+```
+
+Please note that in the example above we created 3 Dask-CUDA workers on one node, for a multi-node cluster, the correct MIG names need to be specified, and they will always be different for each host.
+
+## XGBoost with Dask Cluster
+
+Currently XGBoost only exposes support for GPU communication via NCCL, which does not support MIG. For this reason, A Dask cluster that utilizes XGBoost would have to utilize TCP instead for all communications which will likely cause in considerable performance degradation. Therefore, using XGBoost with MIG is not recommended.
```
jacobtomlinson, line 76: Could we link "XGBoost" to their docs?

source/index.md
```diff
@@ -92,6 +103,7 @@ See our [example notebooks repo](https://github.com/rapidsai/cloud-ml-examples)
 local
 cloud/index
 hpc
+mig
```
jacobtomlinson, line 106: I'm not sure this should be a top-level navigation item. I wonder if we need a new section like "Best practices" or something for topics like this to live under?

jacobtomlinson (commented):
Thanks so much for contributing this here @pentschev this is great!

I've left comments but it's mostly about cross-linking to other documentation.

source/mig.md
pentschev: Done in https://github.com/rapidsai/deployment/pull/63/commits/19c6a4169bc7ae9a262ca2188a82dedf0e3f0ff4 .

source/mig.md
pentschev: Done in https://github.com/rapidsai/deployment/pull/63/commits/19c6a4169bc7ae9a262ca2188a82dedf0e3f0ff4 .

source/mig.md
pentschev: Done in https://github.com/rapidsai/deployment/pull/63/commits/19c6a4169bc7ae9a262ca2188a82dedf0e3f0ff4 .

source/index.md
pentschev: Yeah, I wasn't sure where this would live either, as MIG doesn't accurately fit any of the existing sections. It's technically not a best practice either, since MIG is a specific NVIDIA/CUDA technology. Maybe something like "Other Uses", or something that presents the same idea but better phrased?

source/mig.md
pentschev: Sorry for my ignorance of Sphinx, I see https://github.com/rapidsai/deployment/blob/82e2bb2cf7ede6592449d444faf6d5550818d40a/source/conf.py#L71-L76 but I don't really know how to refer to those, would we need a specific entry for `LocalCUDACluster` there, another for `dask-cuda-worker`, and so on? Or do you mean to simply add a regular link?

jacobtomlinson: No worries! There are neat tricks you can do to cross-link classes and functions from other Sphinx docs. I've added the mapping and updated the mentions of `LocalCUDACluster` to map over in [82ae5c5](https://github.com/rapidsai/deployment/pull/63/commits/82ae5c5b7e5ee97aae1052be889e1851b6832d48)

source/index.md
jacobtomlinson: Yeah that makes sense. I'm tempted to hit merge as is, for now, as I expect we will do a bit of a refactor before 22.12. Other similar sections may come up and inspire the name of a common place to put things.

jacobtomlinson:
Thanks @pentschev!

pentschev:
Thanks @jacobtomlinson for the review and for the intersphinx mapping changes, that is indeed very cool!

PR #74, opened 2022-11-22
Align single node examples

jacobtomlinson:
We now have Single Node examples for AWS, Azure, GCP and IBM Cloud but they were all slightly different.

This PR is a refactor inspired from the best of each page to make them all consistent with each other.

I've also used some includes so that duplicated sections like pulling the Docker image and testing the RAPIDS install is only written once. 

I also tweaked the menus a bit.

### DIFF
--- a/source/_includes/menus/aws.md
+++ b/source/_includes/menus/aws.md
@@ -1,0 +1,45 @@
+`````{grid} 1 2 2 3
+:gutter: 2 2 2 2
+
+````{grid-item-card}
+:link: /cloud/aws/ec2
+:link-type: doc
+Elastic Compute Cloud (EC2)
+^^^
+Launch an EC2 instance and run RAPIDS.
+
+{bdg-primary}`single-node`
+````
+
+````{grid-item-card}
+:link: /cloud/aws/eks
+:link-type: doc
+Elastic Kubernetes Service (EKS)
+^^^
+Launch a RAPIDS cluster on managed Kubernetes.
+
+{bdg-primary}`multi-node`
+````
+
+````{grid-item-card}
+:link: /cloud/aws/ecs
+:link-type: doc
+Elastic Container Service (ECS)
+^^^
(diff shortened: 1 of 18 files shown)


source/_includes/menus/aws.md
```diff
@@ -0,0 +1,45 @@
...
+````
+
+````{grid-item-card}
+:link: /cloud/aws/eks
+:link-type: doc
+Elastic Kubernetes Service (EKS)
+^^^
+Launch a RAPIDS cluster on managed Kubernetes.
+
+{bdg-primary}`multi-node`
```
mroeschke, line 21: Curious on your thoughts of making `multi-node` a different color than `single-node`

source/_includes/menus/gcp.md
```diff
@@ -0,0 +1,24 @@
+`````{grid} 1 2 2 3
+:gutter: 2 2 2 2
+
+````{grid-item-card}
+:link: /cloud/gcp/compute-engine
+:link-type: doc
+Compute Engine Instance
+^^^
+Launch an Compute Engine instance and run RAPIDS.
```
mroeschke, line 9:
```suggestion
Launch a Compute Engine instance and run RAPIDS.
```

mroeschke:
Nice! Learned something new with the VMI for GCP. 

One grammar suggestion and a minor style thing otherwise looks good

source/_includes/menus/aws.md
jacobtomlinson: Yeah I think that would be great. I want to think a bit more deeply about tagging in the docs, especially as we integrate these docs with cloud-ml-examples more closely. So maybe we could have this discussion then?

jacobtomlinson:
Thanks for the review @mroeschke!

mroeschke (approved):

source/_includes/menus/aws.md
mroeschke: Sounds good.

PR #86, opened 2022-12-15
Add SSH instructions to connect to AzureVM

mroeschke:
Let me know if I should hold off on updating the anticipated image tag for nvcr


jacobtomlinson (commented):
In #85 I've updated the `source/_includes/install-rapids-with-docker.md` file to just use Docker Hub. We should just do that here too.

I notice you changes the Ubuntu version to `18` which is going away so that should probably stay as `20`.

mroeschke:
Great, let's just use your update in https://github.com/rapidsai/deployment/pull/85

jacobtomlinson (approved):

PR #87, opened 2022-12-15
Update Vertex AI for 22.12

jacobtomlinson:
Tested the Vertex AI instructions for 22.12. I added the image name explicitly and did some minor style refactoring of the page to make it similar to other pages. Also added a quick cudf example to the bottom to verify things work.


mroeschke (approved):
Nice. Thanks for making the formatting better

Issue #79, opened 2022-12-09
Fix Amazon ECS cluster (dask-cloudprovider) docs


jacobtomlinson:
Test and verify the instructions here https://github.com/rapidsai/deployment/blob/main/source/cloud/aws/ecs.md. Fix up any problems.

PR #88, opened 2022-12-16
Add legacy warning to ECS docs

jacobtomlinson:
The ECS docs here are copied from rapids.ai/cloud and most likely do not work, we have a ticket open to track rewriting it (#79) but it didn't make it into 22.12.

I'm keen to get rapids.ai/cloud redirecting here because the majority of the documentation has been updated and the deployment docs are a much more useful resource than rapids.ai/cloud. So I want to clearly signal that this page is still legacy with a warning and go ahead with the redirect.

image

### DIFF
--- a/source/cloud/aws/ecs.md
+++ b/source/cloud/aws/ecs.md
@@ -1,5 +1,9 @@
 # Elastic Container Service (ECS)
 
+```{warning}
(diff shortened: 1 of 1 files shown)


mmccarty (approved):
Agreed. Let's keep moving forward and prioritize any remaining gaps.

PR #104, opened 2023-01-19
Make GPU test pod instructions easier to maintain

### LINKED ISSUE #90: Make test GPU pod an include
The AWS/GCP/Azure Kubernetes pages all have a common section where you create a Pod that uses a GPU to check that the drivers are installed correctly.

This could be extracted into the `_includes` folder to make updating easier.

xref https://github.com/rapidsai/deployment/pull/84#discussion_r1050138208

betatim:
Closes #90. The instructions have been moved to an "include" file instead of being copy&pasted across multiple pages. This makes it easier to keep the instructions up-to-date and in sync.

The instructions to create the GPU node pool for Azure have been adjusted so that they don't use a taint for the nodes. Without this change we couldn't template the testing instructions because the pod would need to have an additional toleration. I think taints+tolerations make sense in a production setup but we'd probably have to add a sentence or two to explain this to a newcomer. Without the explanation it is probably not super clear what is happening. And for old hands, it is obvious what a taint+toleration is and they can also decide if they want to use it or not in their production setup.

WDYT?


source/cloud/azure/aks.md
```diff
@@ -94,57 +94,19 @@ $ az aks nodepool add \
     --name gpunp \
     --node-count 1 \
     --node-vm-size Standard_NC48ads_A100_v4 \
-    --node-taints sku=gpu:NoSchedule \
```
jacobtomlinson, line 97: I'm totally on board with this change. My only worry is wont there be two node pools now? The default one and the GPU one? So the test pod could be scheduler on either and not work 50% of the time?

Maybe we also need to scale the default node pool down to `0` for the sake of this example?

betatim: (not super familiar with Azure, so 50% guessing) I think there will be two node pools, but I also think that there would be two node pools if using the `--node-taints sku=gpu:NoSchedule`. The thing that will change is that the nodes in the second node pool (called `gpunp`) will no longer have a taint.

The pod used to have a toleration for that particular taint. A toleration makes it possible for it to schedule on a node that has a matching taint. But it doesn't guarantee it will be scheduled there. For that we'd have to use something like node affinity.

Having said all this, I think the Pod will be scheduled on one of the nodes with a GPU, without having to scale down the other pool. This is because the Pod lists a GPU as a resource it requires and the kubernetes scheduler will look for a node where this resource request can be fulfilled.

jacobtomlinson: Yeah you're right! I was thinking about the `gpu:NoSchedule` like a node selector, but it's the inverse. Just stopping the GPU nodes from getting filled up with non-GPU work. The resource requirements will cause it to land on the right node.

PR #111, opened 2023-01-24
Add related examples

### LINKED ISSUE #107: Add related examples section
As described in https://github.com/rapidsai/cloud-ml-examples/issues/197 we need to add a "Related Examples" section to the bottom of each documentation page and crosslink with example notebooks.

image

I expect this will involve creating a custom directive or footer at the bottom of each page that walks the toctree of the examples section. Then walk the notebooks and load the cell metadata tags from the title cell. Then if the notebook tag matches the name of the page, e.g `ec2` a grid item should be generated to link to that notebook.

jacobtomlinson:
Closes #107 

High level changes:
- Adds a Sphinx extension that generates a mapping of tags to notebooks
- Adds a `relatedexamples` directive for use on pages to show related notebooks
- Added tags to the Sagemaker example notebook
- Added the related examples section to the sagemaker page
- ~Added `ruff` for linting~ (Split out into #113)

By adding cell metadata tags to the first cell of a notebook those tags get added to the Sphinx notebook tag mapping. For example I added a `cloud/aws/sagemaker` tag to the Sagemaker notebook.

Then when you use the `relatedexamples` directive on a page it checks for any tags that match the page name. So using it in `source/cloud/aws/sagemaker.md` means it will find the Sagemaker notebook and link to it.

image

There are a few outstanding issues with the PR, but I'm keen to keep moving forwards and open separate issues to track those:
- ~The example link currently uses the doc name rather than the example title because Sphinx hasn't parsed the titles yet at the point I'm processing the directive~ Fixed in 9e37293.
- When you edit a notebook it should mark all the related pages for rebuild, but for some reason, this isn't happening

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 build
 *.swp
+
+__pycache__
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,3 +10,8 @@ repos:
     hooks:
       - id: markdownlint
         args: [--ignore-path=.markdownlintignore]
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    # Ruff version.
+    rev: "v0.0.231"
+    hooks:
+      - id: ruff
--- a/README.md
+++ b/README.md
@@ -2,10 +2,7 @@
 
 ## Building
 
-As a prerequisite, a RAPIDS compatible GPU is required to build the docs since the notebooks in the docs execute the code to generate the HTML output.
-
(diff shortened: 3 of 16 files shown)


betatim (commented):
LGTM, modulo that one comment

betatim:
There is no way to access the site that the CI built right? Just asking because it would be convenient but not required.

jacobtomlinson:
Thanks for the review!

> There is no way to access the site that the CI built right?

Not currently. I opened https://github.com/rapidsai/ops/issues/2413 a while back to ask Ops to set this up but they haven't gotten to it yet.

PR #116, opened 2023-01-25
Add guide on Dask Scheduler hw/sw requirements

### LINKED ISSUE #73: Document Dask scheduler GPU requirements
One question that often comes up is "does the Dask scheduler need a GPU?". Generally, we say "yes" because there are edge cases where objects can be deserialized on the scheduler, but this is pretty anecdotal and maybe not backed up in writing anywhere.

I'd be keen to add an FAQ section to the deployment docs and add this question with a more official answer we can link folks to.

jacobtomlinson:
A common question from users deploying Dask clusters is whether the scheduler has different minimum requirements to the workers. This question is compounded when using RAPIDS and GPUs.

After chatting with @quasiben, @pentschev, @wence- and @rjzamora I've drafted a guide outlining our current advice on scheduler hardware requirements. Given current plans around HLGs this may be subject to change in the future, but wanted to capture the state of today so that we have something to point users to in the short term.

Would appreciate reviews on this to ensure I've captured what we discussed. Also @wence- offered to run some tests to see what various failure modes looked like, it would be nice to update this document with some code examples and errors to help users who may be googling for them.

cc @mmccarty 

Closes #73


source/guides/scheduler-gpu-requirements.md
```diff
@@ -0,0 +1,44 @@
...
+
+If your workload doesn't trigger any edge-cases and you're not using the high-level graph optimizations then you could likely get away with not having a GPU. But it is likely you will run into problems eventually and the failure-modes will be potentially hard to debug.
+
+### Known edge cases
+
+When calling `dask.submit` and passing data directly to a function the whole graph is serialized and sent to the scheduler. In order for the scheduler to figure out what to do with it the graph is deserialized. If the data uses GPUs this can cause the scheduler to import RAPIDS libraries, attempt to instantiate a CUDA context and populate the data into GPU memory. If those libraries are missing and/or there are no GPUs this will cause the scheduler to fail.
+
+Many Dask collections also have a meta object which represents the overall collection but without any data. For example a Dask Dataframe has a meta Pandas Dataframe which has the same meta properties and is used during scheduling. If the underlying data is instead a cuDF Dataframe then the meta object will be too, which can also be deserialized on the scheduler.
```
wence-, line 32: s/can also be/is/

```suggestion
Many Dask collections also have a meta object which represents the overall collection but without any data. For example a Dask Dataframe has a meta Pandas Dataframe which has the same meta properties and is used during scheduling. If the underlying data is instead a cuDF Dataframe then the meta object will be too, which is deserialized on the scheduler.
```

In my testing, this was the thing that actually blew up.

wence- (commented):
A few comments.

I think we actually must be stronger in our statement at the moment.

In almost all workflows it is definitely the case that the scheduler needs a GPU (deserializing device-based `_meta`). But one does not need the same level of resource (e.g. you could get away with a T400, not just a single-GPU box of the same size)

source/guides/scheduler-gpu-requirements.md
```diff
@@ -0,0 +1,44 @@
...
+
+## What does the scheduler use a GPU for?
+
+The Dask client generates a task graph of operations that it wants to be performed and serializes any data that needs to be sent to the workers. The scheduler handles allocating those tasks to the various Dask workers and passes serialized data back and forth. The workers deserialize the data, perform calculations, serialize the result and pass it back.
+
+This can lead users to logically ask if the scheduler needs the same capabilities as the workers/client. It doesn't handle the actual data or do any of the user calculations, it just decides where work should go.
+
+Taking this even further you could even ask "Does the Dask scheduler even need to be written in Python?". Some folks even [experimented with a Rust implementation of the scheduler](https://github.com/It4innovations/rsds) a couple of years ago.
+
+There are two primary reasons why we recommend that the scheduler has the cam capabilities:
```
mmccarty, line 21:
```suggestion
There are two primary reasons why we recommend that the scheduler has the same capabilities:
```

source/guides/scheduler-gpu-requirements.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Known edge cases
+
+When calling `dask.submit` and passing data directly to a function the whole graph is serialized and sent to the scheduler. In order for the scheduler to figure out what to do with it the graph is deserialized. If the data uses GPUs this can cause the scheduler to import RAPIDS libraries, attempt to instantiate a CUDA context and populate the data into GPU memory. If those libraries are missing and/or there are no GPUs this will cause the scheduler to fail.
+
+Many Dask collections also have a meta object which represents the overall collection but without any data. For example a Dask Dataframe has a meta Pandas Dataframe which has the same meta properties and is used during scheduling. If the underlying data is instead a cuDF Dataframe then the meta object will be too, which can also be deserialized on the scheduler.
+
+### Scheduler optimizations and High-Level graphs
+
+The Dask community is actively working on implementing high-level graphs which will both speed up client -> scheduler communication and allow the scheduler to make advanced optmizations such as predicate pushdown.
```
mmccarty, line 36: Link to primary issue for tracking this effort?

source/guides/scheduler-gpu-requirements.md
```diff
@@ -0,0 +1,44 @@
...
+
+Much effort has been put into using traditional serialization strategies to communicate the HLG but this has proven prohibitively difficult to implement. The current plan is to simplify HighLevelGraph/Layer so that the entire HLG can be pickled on the client, sent to the scheduler as a single binary blob, and then unpickled/materialized (HLG->dict) on the scheduler.The problem with this new plan is that the pickle/un-pickle convention will require the scheduler to have the same environment as the client. If any Layer logic also requires a device allocation, then this approach also requires the scheduler to have access to a GPU.
```
mmccarty, line 38:
```suggestion
Much effort has been put into using traditional serialization strategies to communicate the HLG but this has proven prohibitively difficult to implement. The current plan is to simplify HighLevelGraph/Layer so that the entire HLG can be pickled on the client, sent to the scheduler as a single binary blob, and then unpickled/materialized (HLG->dict) on the scheduler. The problem with this new plan is that the pickle/un-pickle convention will require the scheduler to have the same environment as the client. If any Layer logic also requires a device allocation, then this approach also requires the scheduler to have access to a GPU.
```

source/guides/scheduler-gpu-requirements.md
```diff
@@ -0,0 +1,44 @@
...
+
+## So what are the minimum requirements of the scheduler?
+
+From a software perspective we recommend that the Python environment on the client, scheduler and workers all match. Given that the user is expected to ensure the worker has the same environment as the client it is not much of a burden to ensure the scheduler also has the same environment.
+
+From a hardware perspective we recommend that the scheduler has the same capabilities, but not necessarily the same quantity of resource. Therefore if the workers have one or more GPUs we recommend that the scheduler has access to one GPU with matching NVIDIA driver and CUDA versions. In a large multi-node cluster deployment on a cloud platform this may mean the workers are launched on VMs with 8 GPUs and the scheduler is launched on a smaller VM with one GPU. This balance means we can guarantee things function as intended, but reduces cost because placing the scheduler on an 8 GPU node would be a waste of resources.
```
mmccarty, line 44: Does it need to be the same GPU? Could one get away with using T4's on the scheduler and A100's on the workers? Thinking about opportunities for cost savings on Cloud instances.

wence-: I think the answer to this is yes (would be good to test though). The software environment needs to be the same.

source/guides/scheduler-gpu-requirements.md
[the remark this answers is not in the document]
jacobtomlinson: cc @rjzamora who wrote that sentence

wence-:
Here's some more data on what works and doesn't (I can't explore `client.submit` in detail because I have no idea what to do with the result future I get back, it's not obvious to me how to use it in any subsequent computation).

In any case, here's a trivial example:

```python
import cupy
from distributed import Client, wait
import dask.array as da

client = Client(scheduler_file="foo.json")

x = cupy.arange(10)

y = da.arange(1000, like=x)

z = (y * 2).persist()
wait(z)
# Now let's look at some results

print(z[:10].compute())
```

If I run with the `tcp` protocol, this test completes successfully:

```sh
$ CUDA_VISIBLE_DEVICES="" dask scheduler --protocol tcp --scheduler-file foo.json  &
$ dask-cuda-worker --protocol tcp --scheduler-file foo.json &
$ python test.py
2023-01-27 10:58:32,665 - distributed.scheduler - INFO - Receive client connection: Client-89f79995-9e31-11ed-b825-4c77cb2482c7
2023-01-27 10:58:32,665 - distributed.core - INFO - Starting established connection to tcp://192.168.1.167:36646
[ 0  2  4  6  8 10 12 14 16 18]
2023-01-27 10:58:33,959 - distributed.scheduler - INFO - Remove client Client-89f79995-9e31-11ed-b825-4c77cb2482c7
2023-01-27 10:58:33,960 - distributed.core - INFO - Received 'close-stream' from tcp://192.168.1.167:36646; closing.
2023-01-27 10:58:33,960 - distributed.scheduler - INFO - Remove client Client-89f79995-9e31-11ed-b825-4c77cb2482c7
2023-01-27 10:58:33,961 - distributed.scheduler - INFO - Close client connection: Client-89f79995-9e31-11ed-b825-4c77cb2482c7
```

OTOH, if I'm using UCX
```sh
$ CUDA_VISIBLE_DEVICES="" dask scheduler --protocol ucx --scheduler-file foo.json  &
$ dask-cuda-worker --protocol ucx  --scheduler-file foo.json &
$ python test.py
2023-01-27 11:01:27,459 - distributed.scheduler - INFO - Receive client connection: Client-f1d09e8c-9e31-11ed-b919-4c77cb2482c7
2023-01-27 11:01:27,459 - distributed.core - INFO - Starting established connection to ucx://:8786
2023-01-27 11:01:28,263 - distributed.core - ERROR - CUDA error at: .../rmm/include/rmm/cuda_device.hpp:56: cudaErrorNoDevice no CUDA-capable device is detected
Traceback (most recent call last):
  File ".../distributed/distributed/utils.py", line 741, in wrapper
    return await func(*args, **kwargs)
  File ".../distributed/distributed/comm/ucx.py", line 372, in read
    frames = [
  File ".../distributed/distributed/comm/ucx.py", line 373, in <listcomp>
    device_array(each_size) if is_cuda else host_array(each_size)
  File ".../distributed/distributed/comm/ucx.py", line 171, in device_array
    return rmm.DeviceBuffer(size=n)
  File "device_buffer.pyx", line 85, in rmm._lib.device_buffer.DeviceBuffer.__cinit__
RuntimeError: CUDA error at: .../rmm/include/rmm/cuda_device.hpp:56: cudaErrorNoDevice no CUDA-capable device is detected
2023-01-27 11:01:28,263 - distributed.core - ERROR - Exception while handling op gather
Traceback (most recent call last):
  File ".../distributed/distributed/core.py", line 820, in _handle_comm
    result = await result
  File ".../distributed/distributed/scheduler.py", line 5687, in gather
    data, missing_keys, missing_workers = await gather_from_workers(
  File ".../distributed/distributed/utils_comm.py", line 80, in gather_from_workers
    r = await c
  File ".../distributed/distributed/worker.py", line 2872, in get_data_from_worker
    return await retry_operation(_get_data, operation="get_data_from_worker")
  File ".../distributed/distributed/utils_comm.py", line 419, in retry_operation
    return await retry(
  File ".../distributed/distributed/utils_comm.py", line 404, in retry
    return await coro()
  File ".../distributed/distributed/worker.py", line 2852, in _get_data
    response = await send_recv(
  File ".../distributed/distributed/core.py", line 986, in send_recv
    response = await comm.read(deserializers=deserializers)
  File ".../distributed/distributed/utils.py", line 741, in wrapper
    return await func(*args, **kwargs)
  File ".../distributed/distributed/comm/ucx.py", line 372, in read
    frames = [
  File ".../distributed/distributed/comm/ucx.py", line 373, in <listcomp>
    device_array(each_size) if is_cuda else host_array(each_size)
  File ".../distributed/distributed/comm/ucx.py", line 171, in device_array
    return rmm.DeviceBuffer(size=n)
  File "device_buffer.pyx", line 85, in rmm._lib.device_buffer.DeviceBuffer.__cinit__
RuntimeError: CUDA error at: .../rmm/include/rmm/cuda_device.hpp:56: cudaErrorNoDevice no CUDA-capable device is detected
Traceback (most recent call last):
  File "test.py", line 15, in <module>
    print(z[:10].compute())
  File ".../dask/dask/base.py", line 314, in compute
    (result,) = compute(self, traverse=False, **kwargs)
  File ".../dask/dask/base.py", line 599, in compute
    results = schedule(dsk, keys, **kwargs)
  File ".../distributed/distributed/client.py", line 3144, in get
    results = self.gather(packed, asynchronous=asynchronous, direct=direct)
  File ".../distributed/distributed/client.py", line 2313, in gather
    return self.sync(
  File ".../distributed/distributed/utils.py", line 338, in sync
    return sync(
  File ".../distributed/distributed/utils.py", line 405, in sync
    raise exc.with_traceback(tb)
  File ".../distributed/distributed/utils.py", line 378, in f
    result = yield future
  File ".../tornado/gen.py", line 769, in run
    value = future.result()
  File ".../distributed/distributed/client.py", line 2205, in _gather
    response = await future
  File ".../distributed/distributed/client.py", line 2256, in _gather_remote
    response = await retry_operation(self.scheduler.gather, keys=keys)
  File ".../distributed/distributed/utils_comm.py", line 419, in retry_operation
    return await retry(
  File ".../distributed/distributed/utils_comm.py", line 404, in retry
    return await coro()
  File ".../distributed/distributed/core.py", line 1221, in send_recv_from_rpc
    return await send_recv(comm=comm, op=key, **kwargs)
  File ".../distributed/distributed/core.py", line 1011, in send_recv
    raise exc.with_traceback(tb)
  File ".../distributed/distributed/core.py", line 820, in _handle_comm
    result = await result
  File ".../distributed/distributed/scheduler.py", line 5687, in gather
    data, missing_keys, missing_workers = await gather_from_workers(
  File ".../distributed/distributed/utils_comm.py", line 80, in gather_from_workers
    r = await c
  File ".../distributed/distributed/worker.py", line 2872, in get_data_from_worker
    return await retry_operation(_get_data, operation="get_data_from_worker")
  File ".../distributed/distributed/utils_comm.py", line 419, in retry_operation
    return await retry(
  File ".../distributed/distributed/utils_comm.py", line 404, in retry
    return await coro()
  File ".../distributed/distributed/worker.py", line 2852, in _get_data
    response = await send_recv(
  File ".../distributed/distributed/core.py", line 986, in send_recv
    response = await comm.read(deserializers=deserializers)
  File ".../distributed/distributed/utils.py", line 741, in wrapper
    return await func(*args, **kwargs)
  File ".../distributed/distributed/comm/ucx.py", line 372, in read
    frames = [
  File ".../distributed/distributed/comm/ucx.py", line 373, in <listcomp>
    device_array(each_size) if is_cuda else host_array(each_size)
  File ".../distributed/distributed/comm/ucx.py", line 171, in device_array
    return rmm.DeviceBuffer(size=n)
  File "device_buffer.pyx", line 85, in rmm._lib.device_buffer.DeviceBuffer.__cinit__
RuntimeError: CUDA error at: .../rmm/include/rmm/cuda_device.hpp:56: cudaErrorNoDevice no CUDA-capable device is detected
2023-01-27 11:01:28,282 - distributed.scheduler - INFO - Remove client Client-f1d09e8c-9e31-11ed-b919-4c77cb2482c7
2023-01-27 11:01:28,282 - distributed.core - INFO - Received 'close-stream' from ucx://:8786; closing.
2023-01-27 11:01:28,283 - distributed.scheduler - INFO - Remove client Client-f1d09e8c-9e31-11ed-b919-4c77cb2482c7
2023-01-27 11:01:28,284 - distributed.scheduler - INFO - Close client connection: Client-f1d09e8c-9e31-11ed-b919-4c77cb2482c7
```

The same kind of things exhibit with dask-cudf. I think the reason for this is that when UCX is enabled, we can send device buffers without copying to host. Some of the "pass-through" messages that go through the scheduler contain these device buffers which the scheduler needs to allocate a buffer (e.g. the compute call in this example, the result isn't sent direct from workers to client, but comes through the scheduler, which must therefore be cuda-capable to allocate a buffer).

In contrast when using the TCP protocol (not recommended by us for performance reasons if you can avoid it), there is no direct device buffer transfer, so data is first copied to host then pickled, and now the scheduler doesn't fall over.

pentschev:
> The same kind of things exhibit with dask-cudf. I think the reason for this is that when UCX is enabled, we can send device buffers without copying to host. Some of the "pass-through" messages that go through the scheduler contain these device buffers which the scheduler needs to allocate a buffer (e.g. the compute call in this example, the result isn't sent direct from workers to client, but comes through the scheduler, which must therefore be cuda-capable to allocate a buffer).

This is right. I think in theory we could disable the CUDA transport layer in the scheduler, which would then force a similar behavior to that of TCP (device->host->wire->host->device). The logical implication is that this would become a bottleneck which we definitely want to avoid. However, this would be the case in any cloud setups where the scheduler may have a "simple GPU" without any of those capabilities (CUDA IPC nor IB), so this would in any case be a bottleneck, thus simply having a CUDA device on the scheduler may not be sufficient.

source/guides/scheduler-gpu-requirements.md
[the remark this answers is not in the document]
rjzamora: I suppose "existing" would be better than "traditional".  Sorry - that text was created quickly to explain some context in an email, so I didn't take the time to make it very precise.

The basic idea is that, to date, we have been sending the HLG from client to scheduler using the same mechanisms that we use to send the "low-level" (`dict`-based) graph. This means that we manually pickle most things that are needed on the worker, but not the scheduler, and use msgpack via the standard comms system for everything else. It just so happens that this approach becomes quite tricky once you want to run HLG->dict logic on the scheduler, because the graph-materialization logic now needs to work with a mix of serialized and un-serialized objects. It is also tricky to deal with the fact that some data may need to be materialized twice (on client and then on the scheduler).

wence-:
> The logical implication is that this would become a bottleneck which we definitely want to avoid. However, this would be the case in any cloud setups where the scheduler may have a "simple GPU" without any of those capabilities (CUDA IPC nor IB), so this would in any case be a bottleneck, thus simply having a CUDA device on the scheduler may not be sufficient.

I'd like to understand the implications here (so that we can write appropriate recommendations in this guide and beyond).

It is a known performance pitfall (the dask documentation recommends against this too) to move large amounts of data from the client to the workers (and vice versa). So `client.submit/scatter` with large arguments is a Bad Thing™, equally `foo.compute()` for some "large" collection `foo` should be avoided where possible. In some circumstances these patterns will be unavoidable: in those circumstances, having a fancy GPU on the scheduler will be helpful.

In the case where the data we're moving between scheduler and workers/client is "small" (control messages and the like) then we pay an extra latency (DtoH and HtoD) plus we have slightly lower bandwidth in the transfer.

So my (perhaps naive) hope is that the cost of the extra DtoH stuff is not too large, and one perhaps wouldn't notice? I guess we could do some testing.

source/guides/scheduler-gpu-requirements.md
mmccarty: Yeah, I'd like to provide guidance that A100's are not required for the scheduler, if possible.

pentschev:
> In the case where the data we're moving between scheduler and workers/client is "small" (control messages and the like) then we pay an extra latency (DtoH and HtoD) plus we have slightly lower bandwidth in the transfer.
>
> So my (perhaps naive) hope is that the cost of the extra DtoH stuff is not too large, and one perhaps wouldn't notice? I guess we could do some testing.

I guess this will depend on how many GPU transfers will actually occur. Indeed, small control messages may not be too bad if there are relatively few of them, if there are hundreds or thousands for any given task then this might become more of a problem. In any case the main issue is large transfers, that is what I would be mostly concerned with.

source/guides/scheduler-gpu-requirements.md
```diff
@@ -0,0 +1,44 @@
...
+There are two primary reasons why we recommend that the scheduler has the cam capabilities:
+
+- There are edge cases where the scheduler does deserialize data.
+- Some scheduler optimizations require high-level graphs to be pickled on the client and unpickled on the scheduler.
+
+If your workload doesn't trigger any edge-cases and you're not using the high-level graph optimizations then you could likely get away with not having a GPU. But it is likely you will run into problems eventually and the failure-modes will be potentially hard to debug.
+
+### Known edge cases
+
+When calling `dask.submit` and passing data directly to a function the whole graph is serialized and sent to the scheduler. In order for the scheduler to figure out what to do with it the graph is deserialized. If the data uses GPUs this can cause the scheduler to import RAPIDS libraries, attempt to instantiate a CUDA context and populate the data into GPU memory. If those libraries are missing and/or there are no GPUs this will cause the scheduler to fail.
```
jacobtomlinson, line 30:
```suggestion
When calling `client.submit` and passing data directly to a function the whole graph is serialized and sent to the scheduler. In order for the scheduler to figure out what to do with it the graph is deserialized. If the data uses GPUs this can cause the scheduler to import RAPIDS libraries, attempt to instantiate a CUDA context and populate the data into GPU memory. If those libraries are missing and/or there are no GPUs this will cause the scheduler to fail.
```

source/guides/scheduler-gpu-requirements.md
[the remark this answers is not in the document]
jacobtomlinson:
```suggestion
Much effort has been put into using existing serialization strategies to communicate the HLG but this has proven prohibitively difficult to implement. The current plan is to simplify HighLevelGraph/Layer so that the entire HLG can be pickled on the client, sent to the scheduler as a single binary blob, and then unpickled/materialized (HLG->dict) on the scheduler. The problem with this new plan is that the pickle/un-pickle convention will require the scheduler to have the same environment as the client. If any Layer logic also requires a device allocation, then this approach also requires the scheduler to have access to a GPU.
```

source/guides/scheduler-gpu-requirements.md
```diff
@@ -0,0 +1,44 @@
...
+
+A common question from users deploying Dask clusters is whether the scheduler has different minimum requirements to the workers. This question is compounded when using RAPIDS and GPUs.
+
+```{warning}
+This guide outlines our current advice on scheduler hardware requirements, but this may be subject to change.
+```
+
+**TLDR; It is strongly suggested that your Dask scheduler has matching hardware/software capabilities to the other components in your cluster.**
+
+Therefore, if your workers have GPUs and the RAPIDS libraries installed we recommend that your scheduler does too.
```
jacobtomlinson, line 11:
```suggestion
Therefore, if your workers have GPUs and the RAPIDS libraries installed we recommend that your scheduler does too. However the GPU attached to your scheduler doesn't need to be as powerful as the GPUs on your workers, as long as it has the same capabilities and driver/CUDA versions.
```

source/guides/scheduler-gpu-requirements.md
```diff
@@ -0,0 +1,44 @@
...
+
+## So what are the minimum requirements of the scheduler?
+
+From a software perspective we recommend that the Python environment on the client, scheduler and workers all match. Given that the user is expected to ensure the worker has the same environment as the client it is not much of a burden to ensure the scheduler also has the same environment.
+
+From a hardware perspective we recommend that the scheduler has the same capabilities, but not necessarily the same quantity of resource. Therefore if the workers have one or more GPUs we recommend that the scheduler has access to one GPU with matching NVIDIA driver and CUDA versions. In a large multi-node cluster deployment on a cloud platform this may mean the workers are launched on VMs with 8 GPUs and the scheduler is launched on a smaller VM with one GPU. This balance means we can guarantee things function as intended, but reduces cost because placing the scheduler on an 8 GPU node would be a waste of resources.
```
jacobtomlinson, line 44:
```suggestion
From a hardware perspective we recommend that the scheduler has the same capabilities, but not necessarily the same quantity of resource. Therefore if the workers have one or more GPUs we recommend that the scheduler has access to one GPU with matching NVIDIA driver and CUDA versions. In a large multi-node cluster deployment on a cloud platform this may mean the workers are launched on VMs with 8 GPUs and the scheduler is launched on a smaller VM with one GPU. You could also select a less powerful GPU such as those intended for inferencing for your scheduler like a T4, provided it has the same CUDA capabilities, NVIDIA driver version and CUDA/CUDA Toolkit version. 

This balance means we can guarantee things function as intended, but reduces cost because placing the scheduler on an 8 GPU node would be a waste of resources.
```

jacobtomlinson:
Thanks for all the reviews everyone! I think I've addressed most of the feedback so I'm going to merge things here. 

@wence- would you be able to follow this PR up with another that extends the "known edge cases" section to show some of the code examples that you found that cause things to break?

Issue #115, opened 2023-01-25
Add custom toctree card grid directive


jacobtomlinson:
Currently we hide the default toctree directive on index pages and create our own card grid linking to subpages with `sphinx-design` because it looks nicer. 

<img width="1365" alt="image" src="https://user-images.githubusercontent.com/1610850/214550551-7534aef2-b748-4152-8718-507ef0a94d1e.png">

The problem with this is that it means adding the link to new pages in two places and it is easy to forget to add one, especially given that this is not a common practice. The changes to `index.md` in https://github.com/rapidsai/deployment/pull/116/commits/fa6d5eb6bce23ca6b443fee292c827e107af228f are a perfect example of this. It would be easy to get this wrong.

It would be nice to explore whether we could add a small custom sphinx extension that adds a custom directive that generates both for us and removes the duplication. Could the card grid be generated automatically from the toctree?

PR #118, opened 2023-01-27
Add notebook gallery with tag filter

### LINKED ISSUE #106: Create custom examples gallery
As outlined in https://github.com/rapidsai/cloud-ml-examples/issues/197 we need to have a gallery-style page showing all of the example notebooks with their associated tags. We then need to add a filtering sidebar menu to allow users to find the notebooks they are interested in.

image

I expect this involves writing a custom directive that walks the toctree discovering the notebooks. Then walking the notebooks and loading the tags from the first cell metadata. Then render out the grid using the same nodes we are building manually on other pages currently.

For the filtering sidebar we need to drop in a custom sidebar object with all of the tag categories and then write some client-side JavaScript to show/hide grid items depending on which filters are checked.

jacobtomlinson:
Closes #106. With this PR the example notebooks are shown in a gallery and can be filtered by their tags.

image

This change is made up of a few pieces:

- A custom toctree directive which generates a card table using the same notebook card generator from #111 
  - _NB this approach would be useful for closing #115 in a follow up PR_
- A custom sidebar template for the examples index page with filtering checkboxes generated from the tag tree
- Some client-side JavaScript that hides/shows notebook cards based on the filter checkboxes
- Some custom CSS which styles the filter sidebar and colours the tags
- I changed the tag style from `bdg-primary` to `bdg` throughout the site so the custom tag styling would work correctly on all pages

### DIFF
--- a/extensions/rapids_related_examples.py
+++ b/extensions/rapids_related_examples.py
@@ -114,10 +114,6 @@ def build_tag_map(app: Sphinx, env: BuildEnvironment, docnames: list[str]):
 
     env.notebook_tag_map = {}
 
-    # If no pages are being rebuilt skip generating the tag map
-    # if not docnames:
-    #     return
-
     # Build notebook tag map
     for doc in env.found_docs:
         path = app.env.doc2path(doc)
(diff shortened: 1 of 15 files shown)


mroeschke (approved):
Built the docs from this branch and it looks good!

jacobtomlinson:
Thanks for the review!

PR #119, opened 2023-01-27
Add related examples to all pages

jacobtomlinson:
Add related examples directive to all of the relevant pages.

This highlighted a bug where if there are no related examples the build would fail. Fixed that up by returning an empty list if there are no related examples.


mroeschke (approved):

isVoid (approved):

jacobtomlinson:
Thanks for the reviews!

PR #121, opened 2023-01-30
Add notebook tag sidebar

### LINKED ISSUE #109: Add tags to notebooks when rendered
As described in https://github.com/rapidsai/cloud-ml-examples/issues/197 we need to add tags to the notebooks and render them on the page below the title.

We might also want to modify the title section somehow to make it clear that you are viewing an example notebook rather than a regular doc page, and maybe even link to the notebook on GitHub.

I think a good way to tag notebooks would be to ensure that the first cell of each notebook just contains the H1 title element, then add [notebook cell tags](https://jupyterbook.org/en/stable/content/metadata.html) to that cell.

Then we need to somehow plugin to the [myst-nb](https://myst-nb.readthedocs.io/en/latest/configuration.html) renderer and add a section below the first cell that shows the tags and any other information we want.

jacobtomlinson:
Add notebook tags visibly on the notebook page. Closes #109. 

I went with a sidebar for displaying the tags as it was much easier to implement than plugging into the MyST-NB parser.

image

I also updated the examples gallery so that when you check the filter boxes it updates the URL with a query parameter. Then on page load it looks for any tags in the URL and updates the checkboxes initial state. This allows us to link to the examples gallery with filters already checked.

image

I then made the tags on the notebook pages links to the gallery with that tag already filtered.

https://github.com/rapidsai/deployment/blob/ea9dac887a0c164e4fb2450d3f016f2ce3eb84ac/source/_templates/notebooks-tags.html#L6-L8

### DIFF
--- a/extensions/rapids_related_examples.py
+++ b/extensions/rapids_related_examples.py
@@ -144,6 +144,9 @@ def add_notebook_tag_map_to_context(app, pagename, templatename, context, doctre
         except KeyError:
             tag_tree[root] = [suffix]
     context["notebook_tag_tree"] = tag_tree
+    context["notebook_tags"] = [
+        tag for tag, pages in app.env.notebook_tag_map.items() if pagename in pages
+    ]
 
 
 class NotebookGalleryTocTree(TocTree):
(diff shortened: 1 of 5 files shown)


mroeschke (approved):
Nice! I think this is a cool usability feature

jacobtomlinson:
Thanks for the review!

PR #133, opened 2023-02-03
Add optional extended notebook config to enable dask-kubernetes

jacobtomlinson:
Deploying an interactive single-user notebook on Kubernetes can provide a great place to launch further resources. For example, you could install `dask-kubernetes` and use the Dask Operator to create multi-node Dask clusters from your notebooks.

I've added an extra optional section to the Kubernetes Interactive Notebook deployment instructions which shows the extra steps you need to be able to use `dask-kubernetes` from within the notebook.

I used this for a demo at PyData DC last night and it was really neat launching a GKE cluster, running a RAPIDS Notebook pod and then bursting out from that onto an MNMG RAPIDS cluster.

### DIFF
--- a/source/platforms/kubernetes.md
+++ b/source/platforms/kubernetes.md
@@ -46,6 +46,213 @@ spec:
           name: notebook
 ```
 
+````{dropdown} Optional: Extended notebook configuration to enable launching multi-node Dask clusters
+:color: info
+:icon: info
+
+Deploying an interactive single-user notebook can provide a great place to launch further resources. For example you could install `dask-kubernetes` and use the [dask-operator](../tools/kubernetes/dask-operator) to create multi-node Dask clusters from your notebooks.
+
+To do this you'll need to create a couple of extra resources when launching your notebook `Pod`.
+
+### RBAC Service account and role
+
+To be able to interact with the Kubernetes API from within your notebook and create Dask resources you'll need to create a service account with an attached role.
+
+```yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: rapids-dask
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: rapids-dask
+rules:
+  - apiGroups: [""]
+    resources: ["pods", "services"]
+    verbs: ["get", "list", "watch", "create", "delete"]
+  - apiGroups: [""]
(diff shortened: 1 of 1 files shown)


source/platforms/kubernetes.md
```diff
@@ -46,6 +46,213 @@ spec:
...
 
+````{dropdown} Optional: Extended notebook configuration to enable launching multi-node Dask clusters
+:color: info
+:icon: info
+
+Deploying an interactive single-user notebook can provide a great place to launch further resources. For example you could install `dask-kubernetes` and use the [dask-operator](../tools/kubernetes/dask-operator) to create multi-node Dask clusters from your notebooks.
+
+To do this you'll need to create a couple of extra resources when launching your notebook `Pod`.
+
+### RBAC Service account and role
```
mroeschke, line 57: Maybe for myself, what does `RBAC` mean here?

source/platforms/kubernetes.md
```diff
@@ -46,6 +46,213 @@ spec:
...
+        - name: jupyter-server-proxy-config
+          mountPath: /root/.jupyter/jupyter_server_config.py
+          subPath: jupyter_server_config.py
+  volumes:
+    - name: jupyter-server-proxy-config
+      configMap:
+        name: jupyter-server-proxy-config
+```
+
+We also might want to configure Ask to know where to look for the Dashboard via the proxied URL. We can set this via an environment variable in our `Pod`.
```
mroeschke, line 146:
```suggestion
We also might want to configure Dask to know where to look for the Dashboard via the proxied URL. We can set this via an environment variable in our `Pod`.
```

source/platforms/kubernetes.md
jacobtomlinson: Thanks for flagging this acronym as being defined without a definition. It stands for Role Based Acces Control and is the way Kubernetes manages authentication. https://kubernetes.io/docs/reference/access-authn-authz/rbac/

I think it is probably unnecessary in the title anyway so my suggestion would be to just remove it rather than define it.

```suggestion
### Service account and role
```

jacobtomlinson:
Thanks for the review @mroeschke. I think I've addressed the feedback, could you give it another quick look to make sure the auth bit is still clear?

mroeschke (approved):

PR #134, opened 2023-02-07
Style changes to related files and tags

jacobtomlinson:
Minor tweak to related files to ensure directories are sorted at the top and have a trailing slash to show they are a directory. Also changed to a monospace font.

image

Also added a couple of base tag styles for `workflow` and `dataset`.


jacobtomlinson:
Given how minor this is I'm going to go ahead and merge it in.

PR #143, opened 2023-02-09
Add extension and config to enable templating in versions

### LINKED ISSUE #145: Use templating for versions throughout the docs
Currently, we have RAPIDS versions hard-coded throughout our documentation. For each release, we have an activity where we go through all of the pages and update everything. This requires a bunch of PRs to be opened every release and we have to wait until the container images are available so we lag behind the main release.

I'd like to update the docs so that we configure the versions in one place like `source/conf.py` and then template those values in throughout the docs.

One benefit of this is it makes it easy to template in nightly images and versions in nightly builds, and stable images and versions in stable builds. We still need to test that things are working but this means we could move our testing activity into the freeze and run through with the nightly images. We would only have to make PRs for breakages rather than updating every version. 

Then at release time we can bump our stable version in one place and tag a release very quickly after the images and packages become available. This could even be handed off to ops to include in their release process.

jacobtomlinson:
This PR allows any RAPIDS versions for things like container images and install instructions to be configured in `source/conf.py` and then templated into the documentation pages.

```python
# source/conf.py
...

versions = {
    "stable": {
        "rapids_container": "rapidsai/rapidsai-core:22.12-cuda11.5-runtime-ubuntu20.04-py3.9",
    },
    "nightly": {
        "rapids_container": "rapidsai/rapidsai-core-nightly:23.02-cuda11.5-runtime-ubuntu20.04-py3.9",
    },
}

...
```

```markdown
<!-- source/some-doc-page.md -->

# My doc page

The latest container image is {{ rapids_container }}.
```

All builds will use the nightly section by default which allows you to test with the latest and greatest containers when developing locally or previewing nightly docs builds. To build the docs using the stable images you need to set the environment variable `DEPLOYMENT_DOCS_BUILD_STABLE` to `true`. This is done automatically when building from a tag in CI.

```console
$ DEPLOYMENT_DOCS_BUILD_STABLE=true sphinx-autobuild source build/html
```

Before we publish a new version for a release we need to ensure that the latest container images are available and then update the `stable` config to use the new release version and update `nightly` to use the next upcoming nightly. This is a huge improvement over manually having to go through every page and update the RAPIDS version.

The first commit in this PR adds the framework and docs to enable this. The second commit goes through the docs pages and replaces hard coded values with templates.

Closes #145


betatim (approved):
LGTM. I didn't go though all the files where you switched to using the templated (assuming that if you made a typo the build would fail). I looked at a few and they looked reasonable. I was positively surprised that the templating also works for things like notebooks! How cool is that!

Only nit: should the README contain a short section on how to use the templating system? On the one hand it would be useful to spread the knowledge that it exists, on the other hand the README will get longer and longer (and then we end up discussing writing docs for the docs ...).

jacobtomlinson:
> I was positively surprised that the templating also works for things like notebooks! How cool is that!

I know right. I totally expect things to blow up in my face, but so far it hasn't. 

> Only nit: should the README contain a short section on how to use the templating system?

Agree, I've added a small example.

PR #157, opened 2023-02-22
add helper_functions.py file

skirui-source:


source/examples/helper_functions.py
```diff
@@ -0,0 +1,222 @@
+## Copyright (c) 2019-2021, NVIDIA CORPORATION.
```
jacobtomlinson, line 1:
```suggestion
## Copyright (c) 2019-2023, NVIDIA CORPORATION.
```

jacobtomlinson (commented):
Thanks 😁

I think this file needs to be inside the directory with the notebook. Right now it's at the `/source/examples` level. Could you move it inside the example directory that it belongs with?

jacobtomlinson (approved):
🚀

PR #139, opened 2023-02-07
Update to match rapids.ai and doc site refactor

exactlyallan:
Added HPC content (removed from rapids.ai). Updated install links (moved to docs/install). Improved Nav to mach docs site.

NOTE: should sync merge with other updates.

@jacobtomlinson or @mmccarty could you review please?

PR #162, opened 2023-02-22
Improve container docs consistency

jacobtomlinson:
Updating this page to be consistent with the instructions on the Local install page.

Ideally I would like this page to have instructions on building your own images at some point, but this is fine for now.

### DIFF
--- a/source/tools/rapids-docker.md
+++ b/source/tools/rapids-docker.md
@@ -1,6 +1,6 @@
-# Docker Images
+# Container Images
 
-RAPIDS docker images can be found on [Docker Hub](https://hub.docker.com/r/rapidsai/rapidsai).
+Installation instructions for Docker are hosted at the [RAPIDS Container Installation Docs Page](https://docs.rapids.ai/install#docker).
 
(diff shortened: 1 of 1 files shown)


source/tools/rapids-docker.md
```diff
@@ -1,6 +1,6 @@
-# Docker Images
+# Container Images
 
-RAPIDS docker images can be found on [Docker Hub](https://hub.docker.com/r/rapidsai/rapidsai).
+Installation instructions for Docker are hosted at the [RAPIDS Container Installation Docs Page](https://docs.rapids.ai/install#docker).
```
hcho3, line 3: The URL doesn't work ?

jacobtomlinson: Hrm good point. All the links on the `local.md` page seem broken too. @exactlyallan added this in #139, I wonder if that page has moved or something?

jacobtomlinson: Ok spoke to @exactlyallan we should be good to merge, that link will work later today.

PR #167, opened 2023-02-22
[DOC]: Updates dask-operator page with latest package name and fixes the client code

isVoid:
Updates dask-operator page with latest package name and fixes the client code


jacobtomlinson (approved):
Thanks

PR #182, opened 2023-03-17
Include .ipynb files in output with templated values

### LINKED ISSUE #173: docker image name doesn't render correctly in notebook
CC: @jacobtomlinson 

Two issues I discovered when testing the[ hpo notebook ](https://github.com/rapidsai/deployment/blob/main/source/examples/rapids-sagemaker-hpo/notebook.ipynb) for 23.02 release:
- The notebook file from GH doesn't render `" {{ rapids_container }}"` correctly as it does on the docs page. So I had to manually copy/paste  image name 
- TAG names in DockerHub need to be updated: I was having trouble pulling docker image with these tags : `Error response from daemon: manifest for rapidsai/rapidsai:23.02-cuda11.8-runtime-ubuntu18.04-py3.10 not found: manifest unknown: manifest unknown`

      * rapidsai/rapidsai:23.02-cuda11.8-runtime-ubuntu18.04-py3.10
      * rapidsai/rapidsai:23.02-cuda11.5-runtime-ubuntu18.04-py3.10

^^I think it's because 23.02 is only compatible with Ubuntu 20.04 and cuda 11.5 /11.8 (should be updated in DockerHub)

jacobtomlinson:
Closes #173 

Updated the notebook files extension to read in the raw `notebook.ipynb`, pass it through jinja2 to populate the templated values, then write it into the build output alongside the generated `notebook.html`.

Also added a relative link to the related files list so that users can download the populated notebook directly from the docs.

image

I am a little worried that folks will click the other files like `Dockerfile` and be taken to GitHub, and then download the notebook from there which will still have the templates in. We could copy all files into the build output and make them all relative links so that they all download the files and none of them take you to GitHub. I think this concern is orthogonal to this PR though so I'm going to open a new issue to track it.

### DIFF
--- a/extensions/rapids_notebook_files.py
+++ b/extensions/rapids_notebook_files.py
@@ -1,4 +1,10 @@
 import pathlib
+import re
+from functools import partial
+
+
+def template_func(app, match):
+    return app.builder.templates.render_string(match.group(), app.config.rapids_version)
 
 
 def find_notebook_related_files(app, pagename, templatename, context, doctree):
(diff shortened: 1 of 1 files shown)


extensions/rapids_notebook_files.py
```diff
@@ -15,24 +53,32 @@ def find_notebook_related_files(app, pagename, templatename, context, doctree):
...
-        context["related_notebook_files"] = (
-            related_notebook_dirs + related_notebook_files
+        # Copy all related files to output and apply templating
+        related_notebook_files = walk_files(
+            app, path_to_page_parent, path_to_output_parent
         )
 
+        # Make archive of related files
+        if related_notebook_files and len(related_notebook_files) > 1:
+            archive_path = path_to_output_parent / "all_files.zip"
```
betatim, line 68: nit: can we make the filename depend on the example? Maybe `pagename` contains the name of the example? A thing that makes creating a filename from it a bit tricky is that there are many characters that need dealing with when creating a filename, they somehow need to be encoded. So this might be something for a new PR?

jacobtomlinson: Yeah I started to implement this and it was less trivial than I thought so I backed out. Probably something for a follow-up PR as you say.

PR #190, opened 2023-03-22
Change dataset tags to data-storage and data-format

### LINKED ISSUE #188: Dataset tagging scheme in example notebooks
> not sure how I feel about tagging datasets too

_Originally posted by @mroeschke in https://github.com/rapidsai/deployment/pull/185#discussion_r1142636129_

Currently we are adding tags for the datasets we are using. But perhaps it would be more useful for us to be tagging with the dataset format (csv, parquet, etc) and the storage location (S3, GCS, HTTP, etc). I think users are more likely looking for examples of loading data in the same format they use from the same location they store theirs, but probably care less about specific dataset examples.

jacobtomlinson:
Closes #188 

Replaced the `dataset` tags with `data-storage` and `data-format` which will be more useful to users.

I also tweaked the template so that hyphenated tags get shown as two words `data-storage` => `Data Storage`.

image

### DIFF
--- a/source/_static/css/custom.css
+++ b/source/_static/css/custom.css
@@ -105,9 +105,14 @@ div.docref > .admonition-title::after {
   border-left: 0.5em #3194c7 solid;
 }
 
-.tag-dataset {
+.tag-data-format {
   background-color: #cc539d;
   border-left: 0.5em #cc539d solid;
+}
+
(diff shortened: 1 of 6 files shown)


mroeschke (approved):
Nice! I think these tags are definitely more generally descriptive.

PR #194, opened 2023-04-06
Update ECS cluster docs

skirui-source:
Fixes: https://github.com/rapidsai/deployment/issues/79


skirui-source:
Blocked by missing  IAM permission to create the ECS cluster, working with ops team to resolve.

```
There was an error creating cluster rapids-cluster-skirui.
API: autoscaling:CreateAutoScalingGroup You are not authorized to use launch template: lt-0c67a81697a090b5a
```

skirui-source:
> Blocked by missing IAM permission needed to create the ECS cluster, working with ops team to resolve:
> 
> ```
> There was an error creating cluster rapids-cluster-skirui.
> API: autoscaling:CreateAutoScalingGroup You are not authorized to use launch template: lt-0c67a81697a090b5a
> ```

RESOLVED THIS, now blocked by a new missing IAM policy when creating the dask-cluster (step 4):

source/cloud/aws/ecs.md
```diff
@@ -1,70 +1,106 @@
...
 )
 ```
 
-[CLUSTER_ARN] = The ARN of an existing ECS cluster to use for launching tasks <br />
-[NUM_WORKERS] = Number of workers to start on cluster creation. <br />
-[NUM_GPUS] = The number of GPUs to expose to the worker, this must be less than or equal to the number of GPUs in the instance type you selected for the ECS cluster (e.g `1` for `p3.2xlarge`).
+````{note}
+When calling this command for the first time, a security group that matches the cluster name will automatically be created. <br />
+
+However, if the cluster creation fails and you want to use the same cluster for subsequent runs of ECSCluster(), you'll need to provide the `security_groups` parameter with the value of the security group, e.g
```
jacobtomlinson, line 78: If we are suggesting the user creates the security group we should probably mention what rules that security group needs to have.

jacobtomlinson (commented):
Glad to see you got this working. Overall this looks great, I have a few comments about cleaning up formatting and language a little.

source/cloud/aws/ecs.md
skirui-source: It states that the first time you run this command, you do not need to specify the `"security_group" `param as the `ECSCluster()` command will automatically create one for you ...

But if you need to re-use the same ecs cluster for subsequent runs of `ECSCluster() ,` then you will need to provide the SG that was created prior..

 I hope that makes sense, i have updated it.

jacobtomlinson: > But if you need to re-use the same ecs cluster for subsequent runs of ECSCluster() , then you will need to provide the SG that was created prior..

That's interesting. I would expect it to pick up the security group that was created automatically. Perhaps this is a bug in dask-cloudprovier. Anyway thanks for updating here.

source/cloud/aws/ecs.md
```diff
@@ -1,81 +1,104 @@
...
 Create the ECSCluster object in your Python session:
 
 ```python
 from dask_cloudprovider.aws import ECSCluster
 
 cluster = ECSCluster(
-    cluster_arn=[CLUSTER_ARN], n_workers=[NUM_WORKERS], worker_gpu=[NUM_GPUS]
+    cluster_arn= "<cluster arn>",
+    n_workers=<nuum_workers>,
+    worker_gpu=<num_gpus>,
```
jacobtomlinson, line 67:
```suggestion
    n_workers=<num_workers>,
    worker_gpu=<num_gpus>,
```

PR #197, opened 2023-04-12
Add a new page for KServe

hcho3:
Closes https://github.com/rapidsai/cloud-ml-examples/issues/240

For now, document examples with tree models. We should eventually add examples with other RAPIDS models too.


jacobtomlinson (commented):
Looks good thanks. Couple of small comments.

jacobtomlinson (approved):

PR #205, opened 2023-04-12
Tweak Dask CUDA page

### LINKED ISSUE #204: Release 23.04 tasks - guides/cuda/localcluster
[guides/cuda/localcluster](https://docs.rapids.ai/deployment/nightly/guides/cuda/localcluster.html)

jacobtomlinson:
I made a few changes to the `dask-cuda` page including moving it from `guides` to `tools`. Also ensured it is added to a toctree.

Closes #204


skirui-source:
LGTM ! except for my comment above re clean-up.

jacobtomlinson:
I'm not sure your comment saved as I can't see it. But I'm guessing it's related to removing the cleanup section. The `LocalCUDACluster` doesn't really need cleaning up in the same way as other clusters as it only creates subprocesses, so when the Python process exits the other processes will too.

PR #214, opened 2023-04-13
Upload docs to S3

### LINKED ISSUE #168: Copy Documentation files to S3
We are in the process of updating the documentation-building process for RAPIDS libraries so that they upload their HTML files to S3. 

In the future, S3 will be used as the "source of truth" for documentation instead of the https://github.com/rapidsai/docs/ repository. That repository will then be configured to pull the HTML files from S3 during build time.

In preparation for the removal of the HTML files from the `rapidsai/docs` repository, it would be helpful if the HTML files from this repository were also uploaded to S3.

Here are some additional links/comments that may be helpful:

- S3 location where the docs should be pushed (`VERSION_NUMBER` is like `23.04`, `23.06`, etc.): https://github.com/rapidsai/rmm/blob/ebf591ffc2fd7b83605f5725ec6e8ba425858c02/ci/build_docs.sh#L44
- We also made [these sphinx adjustments](https://github.com/rapidsai/rmm/blob/ebf591ffc2fd7b83605f5725ec6e8ba425858c02/ci/build_docs.sh#L38-L39) in order to:
  - generate text docs in addition to HTML docs (this will help Taurean's efforts to create a `diff` tool that will help end-users identify API changes between releases)
  - switch from using `sphinx`'s `html` builder to its `dirhtml` builder. This will generate URL paths that do not include extraneous `.html` extensions (see [sphinx docs](https://www.sphinx-doc.org/en/master/man/sphinx-build.html#options) for more details).

ajschmidt8:
Closes #168.

This PR adjusts the `build-and-deploy.yml` workflow to push the documentation files to S3.

This is needed for the upcoming overhaul to the `rapidsai/docs` repository.

I will open a subsequent PR in the future to remove the step from this workflow that commits files directly to that repository.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -34,6 +34,21 @@ jobs:
           DEPLOYMENT_DOCS_BUILD_STABLE: ${{ startsWith(github.event.ref, 'refs/tags/') && 'true' || 'false' }}
         run: make html
 
+      - uses: aws-actions/configure-aws-credentials@v1-node16
+        if: ${{ github.repository == 'rapidsai/deployment' && github.event_name == 'push' }}
+        with:
+          role-to-assume: ${{ vars.AWS_ROLE_ARN }}
+          aws-region: ${{ vars.AWS_REGION }}
+          role-duration-seconds: 3600 # 1h
+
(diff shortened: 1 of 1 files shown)


jacobtomlinson (approved):
This looks great thanks

PR #219, opened 2023-04-14
Use `dirhtml` builder for `sphinx`

ajschmidt8:
This PR updates the `build-and-deploy.yml` workflow to use the `dirhtml` builder for `sphinx`.

The result of this change is that "pretty" URLs are generated (e.g. `/cloud/ec2` instead of `/cloud/ec2/index.html`).

This follows the convention used by all other RAPIDS libraries (e.g. [cudf](https://github.com/rapidsai/cudf/blob/e6cb2d0c9ed7ffbcaa466d40b952ba4acd8a60ed/ci/build_docs.sh#L36-L47)).


ajschmidt8:
official `sphinx-build` docs for reference: https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-b

Issue #209, opened 2023-04-12
L4 Testing


jacobtomlinson:
Looks like L4 GPUs are [now available on Google Cloud](https://cloud.google.com/blog/products/compute/introducing-g2-vms-with-nvidia-l4-gpus).

It would be great to take these for a spin and make any relevant docs changes.

Some context:
- L4 GPUs replace the older T4 GPUs are are probably going to be most useful for interactive sessions where you want a cheaper GPU and for model serving and inference
- They are in early access some changes may need to be made to your GCP account before you can use them. If so check in with Ops about that.

Tasks:
- [ ] Spin up a VM with L4 instances and try out the RAPIDS container
  - [ ] Run some example notebooks 
  - [ ] Feedback in the deployment channel with any findings
- [ ] **Optionally** Spin up a GKE cluster with L4 nodes and schedule some work on them (if possible currently)

Deliverables:
- [ ] Demo to the rest of the group (or maybe a RAPIDS demo Wednesday)
- [ ] Tweet about it from the RAPIDS account
- [ ] **Optionally** Update deployment docs with references to L4s where it makes sense

mroeschke:
Just noting in our GCP VM docs, the `NVIDIA GPU-Optimized VMI` does not allow for selecting L4 GPUs yet

I created a g2-standard-4 machine with an L4 GPU with their deep learning debian image which has CUDA, then when I tried following https://docs.rapids.ai/deployment/stable/cloud/gcp/compute-engine.html#install-rapids I get

```
(base) mroeschke@l4-test-rapids:~$ docker run --gpus all --rm -it \
>     --shm-size=1g --ulimit memlock=-1 \
>     -p 8888:8888 -p 8787:8787 -p 8786:8786 \
>     rapidsai/rapidsai-core:23.02-cuda11.5-runtime-ubuntu20.04-py3.10
docker: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: Auto-detected mode as 'legacy'
nvidia-container-cli: initialization error: nvml error: driver not loaded: unknown.
```

jacobtomlinson:
What does `nvidia-smi` show on that machine? You might need to use `rapidsai/rapidsai-core:23.02-cuda11.8-runtime-ubuntu22.04-py3.10`.

mroeschke:
[...]

PR #226, opened 2023-04-27
DOC: Add guide configuring GCP VM with L4 GPU

mroeschke:
xref https://github.com/rapidsai/deployment/issues/209

A lot of sections were taken from the existing GCP VM configuration section.

### DIFF
--- a/source/guides/index.md
+++ b/source/guides/index.md
@@ -38,4 +38,14 @@ Guidance on Dask scheduler software and hardware requirements.
 {bdg-primary}`Dask`
 ````
 
+````{grid-item-card}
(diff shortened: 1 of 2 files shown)


jacobtomlinson (commented):
Thanks @mroeschke.

I noticed the "Install RAPIDS" and "Test RAPIDS" sections aren't showing up properly in the build preview. Perhaps the include paths aren't quite right from this location?

mroeschke:
Ah thanks. I fixed the relative path to the sections and they are rendering now.

jacobtomlinson (approved):
Sweet thanks!

PR #233, opened 2023-05-10
BLD: Make Sphinx Warnings errors

mroeschke:
Was able to pick up some references that were not resolving correctly

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -55,6 +55,7 @@ rapids_version["rapids_conda_packages_list"] = rapids_version[
 # ones.
 sys.path.insert(0, os.path.abspath("../extensions"))
 extensions = [
+    "IPython.sphinxext.ipython_console_highlighting",
     "sphinx.ext.intersphinx",
     "myst_nb",
     "sphinxcontrib.mermaid",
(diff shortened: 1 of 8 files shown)


betatim (approved):
I like the "warnings are errors" setting. From other projects I know there are people who aren't fans. What is the feeling amongst the contributors here?

source/conf.py
```diff
@@ -80,6 +81,8 @@
 copybutton_prompt_text = r">>> |\.\.\. |\$ |In \[\d*\]: | {2,5}\.\.\.: | {5,8}: "
 copybutton_prompt_is_regexp = True
 
+suppress_warnings = ["myst.header", "myst.nested_header"]
```
jacobtomlinson, line 84: TIL you can do this. These warnings were why I hadn't bothered with `-W` before.

source/platforms/kubernetes.md
```diff
@@ -2,6 +2,8 @@
 
 RAPIDS integrates with Kubernetes in many ways depending on your use case.
 
+(interactive-notebook)=
```
jacobtomlinson, line 5: Also TIL this is how you do anchors. Thanks!

PR #238, opened 2023-05-24
FIX: Workflow tags and related downloads

### LINKED ISSUE #232: Tag links are not working
Clicking a tag in an example gives a 404.

https://docs.rapids.ai/deployment/stable/examples/rapids-autoscaling-multi-tenant-kubernetes/notebook/

mroeschke:
closes #232

I also notice that the related files download is not working also

* Adds `.DS_Store` to `.gitignore`
* Replaces `make html` with `make dirhtml` to match deployed docs
* Removes `ci/docs/build.sh` (AFAICT the Github Action which is used to deploy is not using this?)


jacobtomlinson (approved):
Thanks @mroeschke! This looks great.

PR #243, opened 2023-06-08
Add note about docker permissions/group

betatim:
When testing https://docs.rapids.ai/deployment/nightly/cloud/gcp/compute-engine/ I had to reconnect my SSH because `groups` didn't list me as being a member of the `docker` group, but `/etc/group` did. I suspect this is because I got added to the group after connecting?

I added this note to let people know that all they need to do is reconnect, not go off and `groupadd` people etc. However I am unsure if it makes sense to have this comment everywhere where the docker commands are included? WDYT?


jacobtomlinson (approved):
LGTM thanks

PR #244, opened 2023-06-08
Update Vertex AI instructions

betatim:
The labels in the UI changed a bit. Otherwise this works.

However, when I start the JupyterLab instance it "never" finishes loading the kernels from the custom image. I waited for 5 or so minutes before giving up :-/ We can add a note about "be patient", because it is confusing that the obvious "start a notebook" button doesn't have `cudf` in it. But I think if you have to wait minutes then that makes it a useless thing to have. Could there be something wrong with the image I built? Or some other explanation for why loading the extra kernels doesn't seem to happen?

Screenshot 2023-06-08 at 10 40 49

### DIFF
--- a/source/cloud/gcp/vertex-ai.md
+++ b/source/cloud/gcp/vertex-ai.md
@@ -24,11 +24,12 @@ $ docker push gcr.io/<project>/<folder>/{{ rapids_container.replace('rapidsai/',
 
 ## Create a New Notebook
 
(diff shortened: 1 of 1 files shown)


jacobtomlinson (commented):
Thanks for making these tweaks.

I agree that you have to wait a long time for the RAPIDS env to sync over, I fell into that trap previously. Do you want to add a note here warning folks about that?

ncclementi:
The current vertex AI page is working, closing this

PR #271, opened 2023-08-09
Conf.py 23.08 update

jarmak-nv:
This PR does two things to `conf.py`

1. Updates the Docker pull commands to match the[ RAPIDS 23.08 updates](https://github.com/rapidsai/docker/issues/539)
2. Updates the script so we only need to update the versions in 1 place to reduce human error possibility

It also updates the versions for `23.08` since we'll need to do that prior to release anyway.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -22,7 +22,7 @@ author = "NVIDIA"
 
 # Single modifiable version for all of the docs - easier for future updates
 stable_version = "23.08"
-nightly_version = "23.10a"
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
Thanks @jarmak-nv. I pushed an extra commit to restore the `-nightly` suffix to `rapids_version`. It looks like we only use that variable in `coiled.md` right now and it makes more sense to name the environment `XX.XX-nightly` so that it is clear when we switch to a stable one.

PR #287, opened 2023-09-23
Update TCO benchmark to run training jobs in parallel

hcho3:
When using VMs with multiple GPUs, it is more advantageous to run multiple training jobs in parallel, rather than running one training job at a time.


jacobtomlinson (approved):
This looks great to me, I'm not an expert here so I am happy to defer to your judgement.

Issue #228, opened 2023-05-02
Document multi-node with Dask on Databricks


jacobtomlinson:
Our [current Databricks documentation](https://docs.rapids.ai/deployment/stable/platforms/databricks/) shows how to launch a Databricks cluster (single and multi-node) with the RAPIDS environment. However it doesn't actually discuss how to leverage the multi-node cluster with Dask.

We need to document this.

~I expect this will involve using [dask-yarn](https://yarn.dask.org/en/latest/index.html) to start the scheduler and worker processes and will need to use `dask-cuda-worker` as their startup script.~ Databricks doesn't use Yarn, we would need to explore whatever it does use.

Related issue https://github.com/rapidsai/deployment/issues/231 covers documenting the single-node setup.

PR #297, opened 2023-11-01
Document multi-node with Dask on Databricks

skirui-source:
Fixes:https://github.com/rapidsai/deployment/issues/228


skirui-source:
cc @jacobtomlinson 
The `init.sh` script below successfully launches MNMG Databricks cluster with DASK scheduler and workers:
```

#!/bin/bash

set -e

echo "DB_IS_DRIVER = $DB_IS_DRIVER"
echo "DB_DRIVER_IP = $DB_DRIVER_IP"

pip install --upgrade pip dask[complete]

if [[ $DB_IS_DRIVER = "TRUE" ]]; then
  echo "This node is the Dask scheduler."
  dask scheduler &
else
  echo "This node is a Dask worker."
  echo "Connecting to Dask scheduler at $DB_DRIVER_IP:8786"
  # Wait for the scheduler to start 
  while ! nc -z $DB_DRIVER_IP 8786; do
    echo "Scheduler not available yet. Waiting..."
    sleep 1
  done
  dask worker tcp://$DB_DRIVER_IP:8786 &
fi
```

^^ Tested with  `13.3 LTS ML (includes Apache Spark 3.4.1, GPU, Scala 2.12)` and `g4dn.12xlarge` instance types (Driver, Worker nodes) 

*******
While trying to launch **dask cuda worker** (`pip install dask-cuda==23.10.0`), we see the error below:
```
databricks fs cat dbfs:/dbfs/databricks/skirui/1024-060157-s1c0mgfg/init_scripts/1024-060157-s1c0mgfg_10_59_251_17/20231102_061645_00_dask_launch_init.sh.stderr.log

ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
ydata-profiling 4.2.0 requires numpy<1.24,>=1.16.0, but you have numpy 1.26.1 which is incompatible.
scipy 1.9.1 requires numpy<1.25.0,>=1.18.5, but you have numpy 1.26.1 which is incompatible.
mleap 0.20.0 requires scikit-learn<0.23.0,>=0.22.0, but you have scikit-learn 1.1.1 which is incompatible.
/databricks/python3/lib/python3.10/site-packages/scipy/__init__.py:155: UserWarning: A NumPy version >=1.18.5 and <1.25.0 is required for this version of SciPy (detected version 1.26.1
  warnings.warn(f"A NumPy version >={np_minversion} and <{np_maxversion}"
/databricks/python3/lib/python3.10/site-packages/dask/cli.py:100: UserWarning: While registering the command with name 'cuda', an exception ocurred; 'function' object has no attribute 'command'.
  warnings.warn(
Usage: dask [OPTIONS] COMMAND [ARGS]...
Try 'dask -h' for help.

Error: No such command 'cuda'.
```

source/platforms/databricks.md
```diff
@@ -1,12 +1,42 @@
...
+#!/bin/bash
+set -e
+
+# The Databricks Python directory isn't on the path in
+# databricksruntime/gpu-tensorflow:cuda11.8 for some reason
+export PATH="/databricks/python/bin:$PATH"
+
+# Install RAPIDS (cudf & dask-cudf) and dask-databricks
+/databricks/python/bin/pip install --extra-index-url=https://pypi.nvidia.com \
+      bokeh==3.2.2 \
```
jacobtomlinson, line 21: I pushed `dask-databricks` `0.2.0` earlier which pins bokeh. So we can remove this from the docs here.

source/platforms/databricks.md
```diff
@@ -1,12 +1,42 @@
...
+      cudf-cu11 \
+      dask[complete] \
+      dask-cudf-cu11 \
+      dask-cuda==23.10.0 \
```
jacobtomlinson, line 25: We probably don't want to hard-code this here, but without the pin we sometimes get `0.18.0` which is a bit weird (@pentschev FYI). We should use the `{rapids_version}` template variable here instead of a hard value.

source/platforms/databricks.md
```diff
@@ -66,6 +88,24 @@ gdf
...
 ```
 
+Connect to the dask client and submit tasks.
+
+```python
+from dask.distributed import Client
+from dask_databricks import DatabricksCluster
+
+cluster = DatabricksCluster()
+client = Client(cluster)
```
jacobtomlinson, line 98: I think we should use the simplified `get_client` version here to reduce information overload.

```suggestion
import dask_databricks

client = dask_databricks.get_client()
```

source/platforms/databricks.md
```diff
@@ -66,6 +88,24 @@ gdf
...
+
+def inc(x):
+    return x + 1
+
+x = client.submit(inc, 10)
+x.result()
+ 11
```
jacobtomlinson, line 105: It would be good to have a cudf example here. Maybe something like

```python
import cudf
import dask

df = dask.datasets.timeseries().map_partitions(cudf.from_pandas)
print(df.x.mean().compute())
```

Or even something better!

jacobtomlinson (commented):
Thanks @skirui-source.

Instead of starting with the MNMG example I think it may be easier to follow if we show the single-node instructions first, then have a second section that walks about multi-node. We don't want users to think that this is Dask-only.

source/platforms/databricks.md
pentschev: My best guess is this is happening because in the early ages we had no upper limit on Dask/Distributed versions in Dask-CUDA, so this probably happens because some other package is pulling a newer Dask version than what the latest `23.10.0` is pinned to. I don't know if/how we could fix this though. @vyasr has a vast knowledge of PyPI wheels so maybe he will have good ideas here.

jacobtomlinson: Could we yank the old releases without the upper limit from PyPI? It's hard to imagine anyone will be using them this many years down the line.

pentschev: It has some historical value, for example in https://developer.nvidia.com/blog/debugging-mixed-python-and-c-language-stack/ we did exactly that, recreated the environment with the proper (old) versions to serve as debugging basis. I agree people should not be using them in a daily basis, but debugging and benchmarking are still valid use cases in my book.

If yanking the packages is the only way forward, I think this is something that should be raised internally for more discussion before we do it to make sure we're not breaking anything else.

jacobtomlinson: I wish we could patch PyPI metadata like conda-forge...

jacobtomlinson (changes requested):
Thanks for addressing the feedback. Looking through the page I think things have been chopped around so much at this point that the instructions don't really make sense. 

Instead of going through and reviewing the doc line-by-line I think it might be more helpful to take a step back and talk about the structure of the page and the goals of each section. We have all the pieces but they aren't independent from each other, and they aren't flowing especially well. I would recommend that once you take another pass that you follow the instructions yourself to make sure they work and make sense and that a novice user would be able to follow them.

There are three independant topics/goals we want to cover on this page:

1. Redirect Spark/RAPIDS users to the Spark-RAPIDS docs site
2. Describe installing RAPIDS libraries on a single-node cluster interactively
3. Describe deploying a multi-node Dask/RAPIDS cluster with `dask-databricks`

We want to assume that each reader is unfamiliar with RAPIDS and GPUs and haven't heard of terms like MNMG, and is probably only trying to do one of these things. 

This means we probably want to break the page down into the following sections.

#### Introduction

Explain to readers that RAPIDS can be used on Databricks to accelerate work/reduce costs in a few different ways. Libraries like cudf can be used on single-node clusters in notebooks, Spark users can install Spark-RAPIDS or alternatively you can install Dask alongside Spark and then use libraries like dask-cudf for multi-node workloads.

#### Spark/RAPIDS

At the moment we have a little section at the bottom about Spark/RAPIDS. Maybe instead we want to move that up to a warning/info box at the very top that says something along the lines of "If you're looking for the RAPIDS Accelerator for Spark click here".

#### Single-node

Describe how to launch a single-node compute cluster using the GPU tensorflow/pytorch container images provided by Databricks. Then show how to install libraries like `cudf` at runtime with `pip` from a notebook. Include an example of using cudf.

#### Multi-node

Describe how to use an init script to install `dask-databricks` and `cudf` and start a GPU Dask cluster with the `--cuda` flag. Show how to connect the client in the notebook. Then include an example of using dask-cudf.

source/platforms/databricks.md
```diff
@@ -63,9 +63,70 @@ gdf
...
+dask databricks run --cuda
 
 ```
 
-## Databricks Spark
+**Note**: To launch a dask cluster with GPU workers, you must parse in `--cuda` flag option when running the command, otherwise the script will launch standard dask workers by default.
+
+### Launch Dask cluster
+
+Once your script is ready, follow the same [instructions](#launch-cluster) to launch a **Multi-node** Databricks cluster.
```
skirui-source, line 107: i tried to link to the instructions from single-node section here, but i'm not sure if it'll reflect on the deployment pages

source/platforms/databricks.md
```diff
@@ -44,15 +44,15 @@ At the time of writing the `databricksruntime/gpu-pytorch:cuda11.8` image does n
...
 !pip install cudf-cu11 dask-cudf-cu11 --extra-index-url=https://pypi.nvidia.com
 !pip install cuml-cu11 --extra-index-url=https://pypi.nvidia.com
 !pip install cugraph-cu11 --extra-index-url=https://pypi.nvidia.com
 ```
 
-### Test Rapids
+### Test RAPIDS
 
 ```python
 import cudf
```
jacobtomlinson, line 58: When I do this I see the following exception.

```pytb
ValueError: pyarrow.lib.IpcWriteOptions size changed, may indicate binary incompatibility. Expected 88 from C header, got 72 from PyObject
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
File <command-393379380291571>, line 1
----> 1 import cudf
      3 gdf = cudf.DataFrame({"a":[1,2,3],"b":[4,5,6]})
      4 gdf

File /local_disk0/.ephemeral_nfs/envs/pythonEnv-dcdef616-1f9f-4d00-8058-b5c9cd579d96/lib/python3.10/site-packages/cudf/__init__.py:19
     16 from rmm.allocators.cupy import rmm_cupy_allocator
     17 from rmm.allocators.numba import RMMNumbaManager
---> 19 from cudf import api, core, datasets, testing
     20 from cudf.api.extensions import (
     21     register_dataframe_accessor,
     22     register_index_accessor,
     23     register_series_accessor,
     24 )
     25 from cudf.api.types import dtype

File /local_disk0/.ephemeral_nfs/envs/pythonEnv-dcdef616-1f9f-4d00-8058-b5c9cd579d96/lib/python3.10/site-packages/cudf/datasets.py:7
      4 import pandas as pd
      6 import cudf
----> 7 from cudf._lib.transform import bools_to_mask
      8 from cudf.core.column_accessor import ColumnAccessor
     10 __all__ = ["timeseries", "randomdata"]

File /local_disk0/.ephemeral_nfs/envs/pythonEnv-dcdef616-1f9f-4d00-8058-b5c9cd579d96/lib/python3.10/site-packages/cudf/_lib/__init__.py:4
      1 # Copyright (c) 2020-2023, NVIDIA CORPORATION.
      2 import numpy as np
----> 4 from . import (
      5     avro,
      6     binaryop,
      7     concat,
      8     copying,
      9     csv,
     10     datetime,
     11     expressions,
     12     filling,
     13     groupby,
     14     hash,
     15     interop,
     16     join,
     17     json,
     18     labeling,
     19     merge,
     20     null_mask,
     21     nvtext,
     22     orc,
     23     parquet,
     24     partitioning,
     25     pylibcudf,
     26     quantiles,
     27     reduce,
     28     replace,
     29     reshape,
     30     rolling,
     31     round,
     32     search,
     33     sort,
     34     stream_compaction,
     35     string_casting,
     36     strings,
     37     strings_udf,
     38     text,
     39     timezone,
     40     transpose,
     41     unary,
     42 )
     44 MAX_COLUMN_SIZE = np.iinfo(np.int32).max
     45 MAX_COLUMN_SIZE_STR = "INT32_MAX"

File avro.pyx:1, in init cudf._lib.avro()

File column.pyx:1, in init cudf._lib.column()

File scalar.pyx:25, in init cudf._lib.scalar()

File interop.pyx:1, in init cudf._lib.interop()

ValueError: pyarrow.lib.IpcWriteOptions size changed, may indicate binary incompatibility. Expected 88 from C header, got 72 from PyObject
```

jacobtomlinson (commented):
I got as far as testing out the single-node instructions but they don't seem to work. When I import `cudf` I get an exception, so I don't think we are installing things correctly.

Did you see the same thing when you tested it?

source/platforms/databricks.md
vyasr: If I'm understanding the problem correctly, instead of using an `==` could you just use a lower bound `dask-cuda >= {rapids_version}`? Or maybe using `=={rapids_version}` is sufficient? I may not be fully understanding the problem we're trying to solve.

source/platforms/databricks.md
skirui-source: Seeing the same error  testing `cudf-cu11` on: 
- `databricksruntime/gpu-tensorflow:cuda11.8` and 
- `databricksruntime/gpu-pytorch:cuda11.8`


The [error ](https://github.com/huggingface/datasets/issues/5923)looks related so i tried downgrade `pyarrow==11.0.0` and upgrade to `pyarrow==14.0.1`:

- `ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
cudf-cu11 23.10.1 requires pyarrow==12.*, but you have pyarrow 14.0.1 which is incompatible.`

skirui-source: I also tried testing with `12.2 LTS ML GPU` and `13.3 LTS ML GPU` runtimes to see if we could bypass the docker containers but fails in both cases: 
```
ImportError: 
================================================================
Failed to import CuPy.

If you installed CuPy via wheels (cupy-cudaXXX or cupy-rocm-X-X), make sure that the package matches with the version of CUDA or ROCm installed.

On Linux, you may need to set LD_LIBRARY_PATH environment variable depending on how you installed CUDA/ROCm.
On Windows, try setting CUDA_PATH environment variable.

Check the Installation Guide for details:
  https://docs.cupy.dev/en/latest/install.html

Original error:
  ImportError: libcudart.so.12: cannot open shared object file: No such file or directory

```

^^ I tried to manually install cuda-toolkit 12  following [instructions](https://developer.nvidia.com/cuda-downloads?target_os=Linux&target_arch=x86_64&Distribution=Ubuntu&target_version=22.04&target_type=deb_network) but it  didn't work

jacobtomlinson: Awesome ok I tested things with `12.2 LTS Runtime` and the `databricksruntime/gpu-tensorflow:cuda11.8` container.

Looking at `nvidia-smi` it shows CUDA 12.2 so I upgraded CUDA toolkit with

```bash
!cd /etc/apt/sources.list.d && \
    mv cuda-ubuntu2204-x86_64.list.disabled cuda-ubuntu2204-x86_64.list && \
    apt-get update && apt-get --no-install-recommends -y install cuda-toolkit-12.2 && \
    mv cuda-ubuntu2204-x86_64.list cuda-ubuntu2204-x86_64.list.disabled
```

Then I installed `cudf-cu12` and `dask-cudf-cu12`.

```python
!pip config set global.extra-index-url https://pypi.nvidia.com
!pip install cudf-cu12 dask-cudf-cu12
```

And things worked as expected!

```
>>> import cudf
>>> gdf = cudf.DataFrame({"a":[1,2,3],"b":[4,5,6]})
>>> gdf
	a	b
0	1	4
1	2	5
2	3	6
```

source/platforms/databricks.md
```diff
@@ -63,9 +63,70 @@ gdf
...
+
+# The Databricks Python directory isn't on the path in
+# databricksruntime/gpu-tensorflow:cuda11.8 for some reason
+export PATH="/databricks/python/bin:$PATH"
+
+# Install RAPIDS (cudf & dask-cudf) and dask-databricks
+/databricks/python/bin/pip install --extra-index-url=https://pypi.nvidia.com \
+      cudf-cu11 \  # installs cudf
+      dask[complete] \
+      dask-cudf-cu11  \ #installs dask-cudf
```
jacobtomlinson, line 94: Sadly we can't add comments after the `\` like this as they get included in the command. My cluster was failing to start until I removed these.

```suggestion
      cudf-cu11 \
      dask[complete] \
      dask-cudf-cu11  \
```

source/platforms/databricks.md
```diff
@@ -63,9 +63,70 @@ gdf
...
+
+Now you should be able to select a "GPU-Accelerated" instance for both **Worker** and **Driver** nodes.
+
+![Screenshot of driver worker node](../images/databricks-worker-driver-node.png)
+
+### Connect to Client
+
+To test RAPIDS, Connect to the dask client and submit tasks.
+
+![Screenshot of dask-client](../images/databricks-mnmg-dask-client.png)
```
jacobtomlinson, line 123: Can you make this a copy/pastable code cell as well as a screenshot?

source/platforms/databricks.md
```diff
@@ -63,9 +63,70 @@ gdf
...
+
+![Screenshot of dask-client](../images/databricks-mnmg-dask-example.png)
```
jacobtomlinson, line 125: I would probably just replace this screenshot with a code cell. Copy/paste makes users lives way easier.

source/platforms/databricks.md
```diff
@@ -63,9 +63,70 @@ gdf
...
+Now you should be able to select a "GPU-Accelerated" instance for both **Worker** and **Driver** nodes.
+
+![Screenshot of driver worker node](../images/databricks-worker-driver-node.png)
+
+### Connect to Client
+
+To test RAPIDS, Connect to the dask client and submit tasks.
+
+![Screenshot of dask-client](../images/databricks-mnmg-dask-client.png)
+
```
jacobtomlinson, line 124: Can you add a little paragraph here that talks about the Dashboard link shown in the screenshot of the client and how folks can access the Dashboard?

jacobtomlinson (commented):
I've done another pass of this, thanks for addressing the feedback. I think the single-node instructions look good now. 

I've gone through and tested the multi-node instructions this morning but my cluster fails to start. It looks like there are some comments breaking the init script. Whenever you make changes to code cells or scripts in documentation you need to make sure you retest again from the beginning.

I've left a couple of other general comments but I think we are 90% of the way there now.

jacobtomlinson:
I pushed 1a555bd which fixes the build failure.

source/platforms/databricks.md
```diff
@@ -63,9 +65,92 @@ gdf
...
+import dask_databricks
+
+
+client = dask_databricks.get_client()
+client
+```
+
+Dask host interactive diagnostic web servers using [Bokeh](https://docs.bokeh.org/en/latest/). The **[Dask dashboard](https://docs.dask.org/en/latest/dashboard.html)**, in particular provides a web-based UI with visualizations and real-time information about the Dask cluster's status i.e task progress, resource utilization, etc.
+
+The Dask dashboard server will start up automatically when the scheduler is created, and is hosted on ports `8087, 8787` by default.
```
jacobtomlinson, line 135: I think we should only mention `8087` here as this is the port that we are proxying in through Databricks. The default is `8787` but Databricks often runs other software including R Shiny which also defaults to `8787` so we changed to a non-standard port to avoid this.

source/platforms/databricks.md
```diff
@@ -63,9 +65,92 @@ gdf
...
+
+```python
+import dask_databricks
+
+
+client = dask_databricks.get_client()
+client
+```
+
+Dask host interactive diagnostic web servers using [Bokeh](https://docs.bokeh.org/en/latest/). The **[Dask dashboard](https://docs.dask.org/en/latest/dashboard.html)**, in particular provides a web-based UI with visualizations and real-time information about the Dask cluster's status i.e task progress, resource utilization, etc.
```
jacobtomlinson, line 133: I don't think we should mention Bokeh here, it's not necessarily useful for users to know that's what we built it with.

```suggestion
The **[Dask dashboard](https://docs.dask.org/en/latest/dashboard.html)** provides a web-based UI with visualizations and real-time information about the Dask cluster's status i.e task progress, resource utilization, etc.
```

jacobtomlinson (commented):
This is looking great! A couple of last nits and then I think this is good to merge.

jacobtomlinson (approved):

Issue #298, opened 2023-11-16
Add a workflow example that uses multi-node Databricks and Dask (ideally also using dask-deltatable)


jacobtomlinson:
```[tasklist]
### Prep tasks
- [x] Choose a workload (xgboost training, cuml training)
- [x] Put a dataset into Delta Lake
- [x] Figure out how to read with dask-deltatable into a dask-cudf dataframe
```

### Workflow structure

Option 1 (goal):
- Start MNMG Dask/RAPIDS cluster on Databricks (what we have in the docs now)
- Read dataset from Delta Lake with `dask-deltatable` into a `dask-cudf` dataframe
- Do some preprocessing
- Train a model
- Summary

Option 2 (stretch goal):
- Start MNMG Dask/RAPIDS cluster on Databricks (what we have in the docs now)
- Generate data somehow in a GPU intensive way (custreamz)
- Append records to an existing Delta Lake table

Related links:
- https://pypi.org/project/dask-deltatable/
- https://blog.dask.org/2023/02/02/easy-cpu-gpu
- https://learn.microsoft.com/en-us/azure/databricks/delta/tutorial
- https://delta.io/blog/2022-10-25-create-delta-lake-tables/
- https://medium.com/rapids-ai/gpu-accelerated-stream-processing-with-rapids-f2b725696a61

PR #300, opened 2023-11-21
Add workflow example for dask databricks 

skirui-source:
Fixes https://github.com/rapidsai/deployment/issues/298


source/examples/xgboost-dask-databricks/notebook.ipynb
```diff
@@ -0,0 +1,2163 @@
+{
```
jacobtomlinson, line 1: <blockquote>The following example need to be run on a machine with at least one NVIDIA GPU.</blockquote><p>Instead of saying we need a GPU we can say something like "Wiith GPUs we can do the following workflow, and we don't even need that GPU locally because Databricks can provide that for us.".</p><p> </p><p>But generally I think here we need to go for a bit more of a cook book style and break things down into a couple of steps:</p><ul><li>dask-databricks can help us launch a Dask cluster alongside Spark on databricks</li><li>Create a new init script that installs RAPIDS and runs dask-databricks</li><li>Create a new multi-node cluster that uses the script</li><li>Once the cluster is running upload this notebook to Databricks and continue running these cells on there</li></ul><p> </p><p>We should also use the [<code>docref</code> admonition](<a href="https://github.com/rapidsai/deployment#docref" target="_blank">https://github.com/rapidsai/deployment#docref</a>) here to better cross-link to the Databricks page. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/deployment/pull/300/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

jacobtomlinson, line 1: <p>I'm curious why this is in a function, and then we just call the function once? If we only need it one time let's just run this in a cell.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/deployment/pull/300/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

jacobtomlinson, line 1: <p>So far we've used Dask to convert from CSV to Parquet, then we are using Spark to Convert Parquert to Delta Lake, then we are ultimately doing to read it back in with Dask.</p><p> </p><p>Maybe we should just cut things down and use dask-deltatable to convert directly from CSV to Delta Lake and skip out Spark altogether?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/deployment/pull/300/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

jacobtomlinson, line 1: <p>These column names look like they are the first line of the data. Do we want to tell it the names when loading? Or even writing?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/deployment/pull/300/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

jacobtomlinson, line 1: <p>Why is this commented? Should we just remove it if we don't want to run it?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/deployment/pull/300/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='35'/>

jacobtomlinson, line 1: <p>We should mention destroyingt the Databricks cluster here again. And maybe also mention that if they just leave it the cluster will time out, which saves them money in case they forget.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/deployment/pull/300/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='40'/>

jacobtomlinson (commented):
Thanks @skirui-source.

I've made a few specific comments in ReviewNB but also would you address the following high-level things:
- Can you add [tags](https://github.com/rapidsai/deployment?tab=readme-ov-file#tags) to the notebook?
- The build is unhappy. I think it doesn't like the notebook cell metadata for some reason. Perhaps because it was saved in Databricks. Could you open an issue to track this and start looking into a solution?
- Could you give it another spelling/grammar pass as I noticed a few things but it's hard to suggest edits on GitHub for notebooks?
- Can you check the consistency on the spelling of things like `Dask/RAPIDS`, `cuDF`, etc.

Thanks!

source/examples/xgboost-dask-databricks/notebook.ipynb
```diff
@@ -0,0 +1,1911 @@
...
```
skirui-source, line 1: <p>I added the tags, including <code>library/dask-databricks</code>and <code>library/dask-deltatable</code> but I was wondering whether we want to document these two packages in the <code>Tools/Packages</code> on the <a href="https://docs.rapids.ai/deployment/stable/tools/" target="_blank">deployment page</a>?      </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/deployment/pull/300/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

jacobtomlinson: I think we cover `dask-databricks` enough on the Databricks platform page, if we made a tools page it would be much of the same content. I don't think we need to document `dask-deltatable` because it's not a deployment tool.

source/examples/xgboost-dask-databricks/notebook.ipynb
```diff
@@ -0,0 +1,1881 @@
...
```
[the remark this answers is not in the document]
jacobtomlinson: <p>This was marked as resolved but none of these links have been added.</p>

jacobtomlinson (changes requested):
I looks like one comment was marked as resolved but not actually completed. Could you look at adding some more links throughout? Generally whenever we mention a technology like Dask, XGBoost, etc for the first time in a page we should link to it.

Also it looks like the build is still failing

```
/home/runner/work/deployment/deployment/source/examples/xgboost-dask-databricks/notebook.ipynb:40002: WARNING: skipping unknown output mime type: application/vnd.databricks.v1+bamboolib_hint [mystnb.unknown_mime_type]
```

Could you fix that up? Thanks!

jacobtomlinson (approved):
I removed the metadata that sphinx was complaining about.

PR #318, opened 2024-01-17
Add `cudf.pandas`  notebook example for Databricks

skirui-source:
This PR adds existing [workflow example](https://colab.research.google.com/drive/12tCzP94zFG2BRduACucn5Q_OcX1TUKY3) using `cudf.pandas` to the deployment docs. 

This notebook has been tested in Databricks.


jacobtomlinson (changes requested):
I think adding a `cudf.pandas` on Databricks example is a great idea, but I have a few concerns about this PR.

This notebook is identical to the regular 10 mins to `cudf.pandas` example. 

That's problematic for a couple of reasons:
- Search engines generally downweight duplicated content.
- It's not clear which is the canonical source for this content.
- We now have to try and keep things in sync.
- There is nothing here specific to Databricks.

Here are a few things I would change here:
- We should trim this example right down, perhaps just down to the CSV read. 
- It would also be good to talk a bit about Databricks and how it is common for folks to use Pandas in Databricks Notebooks and how adding `cudf.pandas` will benefit them.
- We should link back to the full example at the end if they want to learn more about `cudf.pandas` and to give them a clear direction to go next.

skirui-source:
CC @jacobtomlinson 
> * We should trim this example right down, perhaps just down to the CSV read.
> * It would also be good to talk a bit about Databricks and how it is common for folks to use Pandas in Databricks Notebooks and how adding `cudf.pandas` will benefit them.
> * We should link back to the full example at the end if they want to learn more about `cudf.pandas` and to give them a clear direction to go next.

In response to this, since the notebook isn't specific / relevant to Databricks deployment , we can just  avoid documenting it  altogether and instead do this: 

-  Under the **'Test RAPIDS'** section of the Single-node instructions, in addition to the simple `import cudf` example, i can talk about the benefits of substituting pandas with `cudf.pandas` and link the [original coLab notebook](https://colab.research.google.com/drive/12tCzP94zFG2BRduACucn5Q_OcX1TUKY3) as a  workflow example 
- I think that's a better alternative to trimming down the original notebook 

Thoughts?

Issue #299, opened 2023-11-17
Upgrade to CUDA `12.2` in Databricks ML Runtimes


skirui-source:
CC: @jacobtomlinson

jacobtomlinson:
I think generally we want to figure out the most ergonomic route to CUDA 12 on Databricks.

Right now when you spin up a node you get CUDA 12.2 with the NVIDIA Drivers on Databricks. However, the Databricks container we are instructing folks to use has CUDA Toolkit 11.8 so you need to install `cudf-cu11`.

It is possible to upgrade CUDA Toolkit either via an init script or at runtime, but it adds some complexity.

It would also be interesting to explore avoiding using the containers altogether and using the ML Runtime and see what CUDA Toolkit and other versions you get with that. But I tried doing this yesterday and couldn't get to a working RAPIDS setup.

skirui-source:
@jacobtomlinson I have tested rapids installation in multi-node Databricks cluster with this init script: 

```
#!/bin/bash
set -e

# Install RAPIDS libraries
pip install \
    --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple \
    "cudf-cu12>=24.02.0a0,<=24.02" "dask-cudf-cu12>=24.02.0a0,<=24.02" \
    "cuml-cu12>=24.02.0a0,<=24.02" "dask-cuda>=24.02.0a0,<=24.02"
```


Successfully worked for the following Databricks ML Runtimes, (installed  Driver Version: 535.54.03  && CUDA Version: 12.2)
```
- 13.3 LTS ML (GPU, Scala 2.12, Spark 3.4.1)
- 14.0, 14.1, 14.2 LTS ML (GPU, Scala 2.12, Spark 3.5.0)

```
 Failed:
```
 - 12.2 LTS ML (GPU, Scala 2.12, Spark 3.3.2)
 - 14.3 LTS ML Beta(GPU, Scala 2.12, Spark 3.5.0)
```

skirui-source:
Actually for the multi-node instructions, it looks like the init script was successful in launching a Dask cluster (installed rapids, dask, dask-databricks etc) but the dask client fails to connect `TimeoutError`

```
OSError: Timed out trying to connect to tcp://10.59.171.113:8786 after 30 s
---------------------------------------------------------------------------
ConnectionRefusedError                    Traceback (most recent call last)
ConnectionRefusedError: [Errno 111] Connection refused

The above exception was the direct cause of the following exception:

[...]

PR #324, opened 2024-01-24
Update docs to install RAPIDS in ML Databricks Runtimes (via init scripts) 

skirui-source:
Fixes: https://github.com/rapidsai/deployment/issues/299

Right now when you spin up a node you get CUDA 12.2 with the NVIDIA Drivers on Databricks. However, the Databricks container we are instructing  to users has CUDA Toolkit 11.8  so you need to install `cudf-cu11`.

After testing, I was able to successfully pip install RAPIDS via init script using the  ML Runtimes. This PR will update the docs, to avoid using the `tensorflow/pytorch `docker runtime containers altogether. 

Successfully worked for the following Databricks ML Runtimes, (installed Driver Version: 535.54.03 && CUDA Version: 12.2)
```
- 13.3 LTS ML (GPU, Scala 2.12, Spark 3.4.1)
- 14.0, 14.1, 14.2 ML (GPU, Scala 2.12, Spark 3.5.0)
```

Failed:
```
 - 12.2 LTS ML (GPU, Scala 2.12, Spark 3.3.2)
 - 14.3 LTS ML Beta(GPU, Scala 2.12, Spark 3.5.0)
```


source/platforms/databricks.md
```diff
@@ -8,6 +8,29 @@ You can install RAPIDS on Databricks in a few different ways:
...
+#!/bin/bash
+set -e
+
+# Install RAPIDS libraries
+/databricks/python/bin/pip install \
+    --extra-index-url=https://pypi.nvidia.com \
+    cudf-cu12 \
+    dask-cudf-cu12 \
+    cuml-cu12 \
+    dask-cuda=={rapids_version}
```
jacobtomlinson, line 31:
```suggestion
    dask-cuda=={{rapids_version}}
```

source/platforms/databricks.md
```diff
@@ -20,42 +43,20 @@ In order to launch a GPU node uncheck **Use Photon Acceleration**.
 
 ![Screenshot of Use Photon Acceleration unchecked](../images/databricks-deselect-photon.png)
```
jacobtomlinson, line 44: It looks like we missed a step here where we tell folks which Databricks Runtime to choose. Can you add an extra paragraph with this information?

jacobtomlinson (changes requested):
This is awesome! Thanks for pushing on this.

A few high level thoughts:
- I noticed that when choosing the `13.3 ML LTS` Tensorflow is broken by the RAPIDS install. On the `14.2 ML` image it works fine because it has a more recent TF version. Maybe we should only suggest people use `14.2 ML` for now.
- Our init scripts are relatively simple, but creating them adds some extra complexity. I wonder if we can host some example scripts in a public S3 bucket and just tell folks to use this unless they explicitly want to customize them.

skirui-source:
Update: 

For both single and multi-node clusters, I am able to successfully pip install (and import) Rapids and other libraries via init script with the **14.2 LTS ML Runtime (GPU, Scala 2.12, Spark 3.5.0)**.

However,  with the same ML Runtime, i experience issues launching a multi-node Dask cluster with `dask-databricks`, i.e

-  `dask databricks run` : the cluster successfully launches and connects to dask scheduler and workers but fails to submit tasks to the client : `RuntimeError: CuPy failed to load libnvrtc.so.12: OSError: libnvrtc.so.12: cannot open shared object file: No such file or directory`

- `dask databricks run --cuda` completely fails to launch the cluster. The init script logs show : `OSError: libnvidia-ml.so.1: cannot open shared object file: No such file or directory` and `pynvml.nvml.NVMLError_LibraryNotFound: NVML Shared Library Not Found`

skirui-source:
@jacobtomlinson , did you want the init scripts stored in public S3 bucket (i filed ops ticket) or should we leave as is? 

Also, Ready for Review

jacobtomlinson:
Let's open a separate ticket for putting a copy in S3 to simplify things (would you mind opening that?). Then we can get this merged.

skirui-source:
i have already filed [issue ](https://github.com/rapidsai/ops/issues/3046)with ops team to request  a copy in S3 bucket. I think we can go ahead and merge this unless you have any additional feedback?

jacobtomlinson (approved):

PR #349, opened 2024-03-30
Update ec2.md with note on failing to install drivers

raybellwaves:
Just came across this error while connecting to an EC2 instance. The fix was suggested at https://forums.developer.nvidia.com/t/nvidia-gpu-optimized-ami-is-missing-drivers-and-cant-run-the-pytorch-ngc-dockerfile/276492/2 and made a note here in case others come across this.

### DIFF
--- a/source/cloud/aws/ec2.md
+++ b/source/cloud/aws/ec2.md
@@ -28,6 +28,10 @@ Next we need to connect to the instance.
 
 ```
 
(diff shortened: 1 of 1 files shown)


jacobtomlinson (approved):

jacobtomlinson:
Thanks @raybellwaves

PR #380, opened 2024-06-04
add missing imports for dask-cuda docs

jameslamb:
Contributes to #374 .

Fixes the following in `dask-cuda` docs at https://docs.rapids.ai/deployment/nightly/tools/dask-cuda/:

* marks Python notebook cells as Python code, so that:
    - they'll be Python syntax-highlighted when rendered
    - linters can cover them in CI
* adds missing imports
    - *it isn't obvious from this page that `LocalCUDACluster` is something you import from `dask_cuda`*
* removes `if __name__ == "__main__"` stuff in example
    - so that the example code is runnable in a notebook or REPL without the extra friction of writing it to a script


jacobtomlinson (approved):
These changes are a nice improvement.

PR #396, opened 2024-07-22
update versions of dependencies

jameslamb:
In preparation for the 24.08 release, proposes updating various versions across the repo.

* updating `pre-commit` hooks via `pre-commit autodupate`
* putting floors on all dependencies in the docs-building conda environment
  - *(to make `conda` solves faster, and to prevent weird solves from accidentally rolling us back to, for example, much older versions of `sphinx`)*
* using `os: "ubuntu-lts-latest"` and `python: "mambaforge-latest"` readthedocs configs
  - *to avoid the need to make any manual changes here when Ubuntu or `mambaforge` versions go end-of-life*
  - *docs: https://docs.readthedocs.io/en/stable/config-file/v2.html#build-os*
* updating various third-party GitHub Actions to their latest versions

## Benefits of these changes

Makes CI a bit faster.

Improves the likelihood of `pre-commit` hooks catching issues.

Resolves these warnings from GitHub Actions:

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v2, actions/setup-python@v2, pre-commit/action@v2.0.0. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

> The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2, actions/setup-python@v2, pre-commit/action@v2.0.0. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

([example build where you can see the warnings](https://github.com/rapidsai/deployment/actions/runs/9678951985))

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -26,7 +26,7 @@ jobs:
           fetch-depth: 0
 
       - name: Set up Python
-        uses: conda-incubator/setup-miniconda@v2
+        uses: conda-incubator/setup-miniconda@v3
         with:
           miniforge-variant: Mambaforge
           use-mamba: true
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -9,8 +9,8 @@ jobs:
     name: "pre-commit hooks"
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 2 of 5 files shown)


jacobtomlinson (approved):
Looks great, thanks for picking up the hygiene work here!

Issue #377, opened 2024-06-04
Start defaulting to CUDA 12 and Python 3.12 in examples


jameslamb:
## Description

RAPIDS supports CUDA 11.x and 12.x, and has for a couple of releases.

Throughout the examples in this repo, most examples default to CUDA 11. For example:

https://github.com/rapidsai/deployment/blob/26ae5ec9a98f54bfd36362bbb02159064c43717c/source/conf.py#L27-L42

We should update the defaults in these docs to encourage the use of CUDA 12.

## Benefits of this work

* nudges RAPIDS users to move to newer CUDA versions, with all the performance, usability, and other improvements that comes with

## Acceptance Criteria

* all docs published from this repo use CUDA 12 in examples recommending container images and packages, except those where this is a compelling and specific reason to stay on CUDA 11

## Approach

Start from here:

https://github.com/rapidsai/deployment/blob/26ae5ec9a98f54bfd36362bbb02159064c43717c/source/conf.py#L27-L42

And then search through the code for other CUDA 11 and CUDA 12 references. Replace them and test the new references.

## Notes

N/A

raydouglass:
I'd say to also update Python to 3.11. It would be good to always promote the latest versions if possible.

jakirkham:
This is now being done in the nightlies: https://github.com/rapidsai/deployment/pull/398

jameslamb:
RAPIDS has been supporting Python 3.12 for a couple months now, so I've updated this from `3.11 -> 3.12`

jameslamb:
This is now complete (for more details, see https://github.com/rapidsai/deployment/issues/382#issuecomment-2616548127).

The default configurations are all using CUDA 12 and Python 3.12:

https://github.com/rapidsai/deployment/blob/86f77ff0091a4617ab074f9ed26c01316cce0fd1/source/conf.py#L27-L48

PR #398, opened 2024-07-24
start using CUDA 12.5, Python 3.11 images in nightly docs

jameslamb:
Contributes to #377.

Proposes using CUDA 12.5, Python 3.11 container images for the nightly docs. This should make it easier to test with these versions during release testing for 24.08. And hopefully that testing goes well and we can move on to setting the stable versions to these images as well.

CUDA 12.5 `rapidsai/base` and `rapidsai/notebooks` images were added in https://github.com/rapidsai/docker/pull/689, as part of https://github.com/rapidsai/build-planning/issues/73.


jakirkham:
Thanks James! 🙏

Seems like a good idea 👍

Let's see what @jacobtomlinson thinks 🙂

jacobtomlinson (commented):
This seems like a good change.

I just want to double check my understanding about backward compatibility. If a platform gives us nodes with CUDA 12.0, are we fine to use the CUDA 12.5 images? My understanding is that there is backward compatibility within major versions these days, but I just want to check that is true.

jakirkham:
Yes that is still how things work

As long as they supply hardware that supports CUDA 12 and there is a driver that supports CUDA 12 (either provided by them or us via a container), then it should work

Please see [this page for driver versions]( https://docs.nvidia.com/deploy/cuda-compatibility/index.html#cuda-11-and-later-defaults-to-minor-version-compatibility )

jacobtomlinson:
Thanks @jakirkham that's really helpful!

jakirkham:
Thanks Jacob and James! 🙏

PR #403, opened 2024-08-06
point to new dask-cuda docs, ensure nightly deployment docs point to nightly RAPIDS API docs

jameslamb:
Contributes to #402.

`dask-cuda` docs were moved from readthedocs to `docs.rapids.ai` over a year ago (https://github.com/rapidsai/dask-cuda/pull/1211).

This removes the remaining links to that readthedocs site, in favor of directing to the new official home of `dask-cuda` docs.

While doing this, I also noticed that there are many places with hard-coded `/stable` or `/nightly` references to `docs.rapids.ai/api`. This proposes using templating to ensure that on the `/nightly` version of the deployment docs, those point to the `/nightly` version of RAPIDS docs.

* stable: https://docs.rapids.ai/api/dask-cuda/stable/
* nightly: https://docs.rapids.ai/api/dask-cuda/nightly/

cc @charlesbluca for awareness


source/conf.py
```diff
@@ -27,13 +27,15 @@
 versions = {
     "stable": {
         "rapids_version": stable_version,
+        "rapids_api_docs_version": stable_version,
```
jameslamb, line 30: @jacobtomlinson does this templating make sense?

Right now these docs have a mix of hard-coded `/stable` or `/nightly` in their `docs.rapids.ai/api` links. For example:

https://github.com/rapidsai/deployment/blob/84c6375469d93c4e59229ba881d6911dbcba510e/source/platforms/kubeflow.md?plain=1#L70

https://github.com/rapidsai/deployment/blob/84c6375469d93c4e59229ba881d6911dbcba510e/source/platforms/databricks.md#L107

I was thinking that that could mean that if you're looking at the 24.04 deployment docs, you'd possibly see links to the `/stable` docs that take you to, say, `cuml`'s 24.08 site.

But now that I look around... is it even possible to look at the deployment docs for old releases?

Or do we only publish to these two places?

```text
https://docs.rapids.ai/deployment/stable/
https://docs.rapids.ai/deployment/nightly
```

source/tools/kubernetes/dask-helm-chart.md
```diff
@@ -50,7 +50,7 @@ This ensures that dask scheduler and worker versions match.
 `[jupyter|worker].resources` explicitly requests a GPU for each worker pod and the Jupyter pod, required by many accelerated libraries in RAPIDS.
 
 `worker.dask_worker` is the launch command for dask worker inside worker pod.
-To leverage the GPUs assigned to each Pod the [`dask_cuda_worker`](https://docs.rapids.ai/api/dask-cuda/stable/index.html) command is launched in place of the regular `dask_worker`.
+To leverage the GPUs assigned to each Pod the [`dask_cuda_worker`](https://docs.rapids.ai/api/dask-cuda/{{rapids_api_docs_version}}/index.html) command is launched in place of the regular `dask_worker`.
```
jacobtomlinson, line 53: Unfortunately if you look at the build preview this link seems to be broken.

https://rapids-deployment--403.org.readthedocs.build/en/403/tools/kubernetes/dask-helm-chart/

It takes you to https://docs.rapids.ai/api/dask-cuda/%7B%7Brapids_api_docs_version%7D%7D/index.html

jacobtomlinson (changes requested):
I think template values need spaces around them, like `{{ rapids_api_docs_version }}`.

So the template values here need updating.

jameslamb:
hmmm yeah, so I guess this:

https://github.com/rapidsai/deployment/blob/6eeb01ab8b451fb5d197917b997796281c718562/extensions/rapids_version_templating.py#L13

must run on the generated HTML, after it's already been rendered from markdown?

Because this is what I see in the page source for view-source:https://rapids-deployment--403.org.readthedocs.build/en/403/tools/kubernetes/dask-helm-chart/

```html
[<code class="docutils literal notranslate"><span class="pre">dask_cuda_worker</span></code>](https://docs.rapids.ai/api/dask-cuda/nightly/index.html)
```

Would it be too weird to introduce some other, valid-in-a-URL, templating character for these cases? Like this?

```text
https://docs.rapids.ai/api/dask-cuda/~~~rapids_api_docs_version~~~/index.html
```

And then modifying the replacement code to also replace that? I'm gonna try that.

jacobtomlinson:
I like that idea! But maybe it looks like that implementation still needs the spaces.

jameslamb:
Got pulled away from this near the end of the day, it is not ready to merge. I'll get the templating working tomorrow.

jacobtomlinson:
No worries, I've moved it back to draft. Let me know when it's ready.

source/tools/kubernetes/dask-helm-chart.md
jameslamb: this is working now 😁 

https://rapids-deployment--403.org.readthedocs.build/en/403/tools/kubernetes/dask-helm-chart/

jameslamb:
Ok @jacobtomlinson , I think I got this working!

As it turns out, none of these were correct:

* "the templating requires spaces inside the template characters"
* "the templating must run on the already-generated HTML"

These URLs weren't being templated because **the `docutils` code in this project was not processing URLs at all**!

The use of `docutils.nodes.SparseNodeVisitor` here...

https://github.com/rapidsai/deployment/blob/b6c222f8c29de49a14ebc7157c49b5d863a29983/extensions/rapids_version_templating.py#L6

... tells `docutils` "only visit document nodes for which the visitor has a corresponding `visit_{node_class}()` method defined". In our case here, that meant only `Text` nodes, because that was the only `visit_{node_class}()` method defined:

https://github.com/rapidsai/deployment/blob/b6c222f8c29de49a14ebc7157c49b5d863a29983/extensions/rapids_version_templating.py#L11

`docutils` stored links (like those that in HTML would be `<a href="..."></a>`) as a `docutils.nodes.reference` instance. I added a `visit_reference()` method on the class doing the templating, and implemented that method in a way that works with how `docutils` stores URLs.

Other changes I made, to hopefully make this easier for the next person:

* added docstrings, links, and type hints throughout the templating code
* renamed `TextNodeVisitor -> RapidsCustomNodeVisitor`, to make it clearer that that name doesn't affect anything about how `docutils` chooses which nodes to process

I think we still want the use of a URL-valid templating string like `~~~, so that:

* `docutils` will correctly parse those links as `docutils.nodes.reference` objects
* `pre-commit` hooks that lint markdown will not complain about bare URLs

extensions/rapids_version_templating.py
```diff
@@ -1,29 +1,89 @@
...
+from copy import deepcopy
+from typing import TYPE_CHECKING
 
 from docutils import nodes
 
+if TYPE_CHECKING:
+    import sphinx
+
 
 class TextNodeVisitor(nodes.SparseNodeVisitor):
```
jacobtomlinson, line 11: You mentioned renaming this, does this still need to be done?

source/hpc.md
```diff
@@ -50,7 +50,7 @@ The scheduler doesn't need the whole node to itself so we can also start a worke
 
 ### Start Dask CUDA Workers
 
-Next start the other [dask-cuda workers](https://dask-cuda.readthedocs.io/). Dask-CUDA extends the traditional Dask `Worker` class with specific options and enhancements for GPU environments. Unlike the scheduler and client, the workers script should be _scalable_ and allow the users to tune how many workers are created.
+Next start the other [dask-cuda workers](https://docs.rapids.ai/api/dask-cuda/~~~rapids_api_docs_version~~~/). Dask-CUDA extends the traditional Dask `Worker` class with specific options and enhancements for GPU environments. Unlike the scheduler and client, the workers script should be \_scalable\* and allow the users to tune how many workers are created.
```
jacobtomlinson, line 53: Was the `\_scalable\*` intentional here? It [renders as `_scalable*` in the preview](https://rapids-deployment--403.org.readthedocs.build/en/403/hpc/#start-dask-cuda-workers).

jacobtomlinson (commented):
Thanks for taking the time to dive into this. Just a couple of small comments then this is good to go!

extensions/rapids_version_templating.py
jameslamb: yes definitely, I must hav written that in the description and then got distracted before pushing it, sorry 😅 

Will do that shortly.

jameslamb: updated! sorry about that

source/hpc.md
jameslamb: This was definitely **not** intentional, thanks for catching it.

When I was experimenting with other templating approaches and tested something like this:

```text
[dask-cuda workers](https://docs.rapids.ai/api/dask-cuda/{{ rapids_api_docs_version }}/
```

That space character and the use of `{{` together led one of the precommit hooks (I think `markdownlint`? I don't remember) to make some weird decisions about where that URL ended. And it added this weird escaping as a result.

Now that we're using template characters that are valid for URLs, and without spaces, that's not a problem 😁

jacobtomlinson (approved):

PR #411, opened 2024-08-08
Update to 24.08

raydouglass:
Update the versions for RAPIDS `24.08` release

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -21,8 +21,8 @@ copyright = f"{datetime.date.today().year}, NVIDIA"
 author = "NVIDIA"
 
 # Single modifiable version for all of the docs - easier for future updates
-stable_version = "24.06"
-nightly_version = "24.08"
+stable_version = "24.08"
+nightly_version = "24.10"
 
 versions = {
     "stable": {


source/conf.py
```diff
@@ -21,8 +21,8 @@
 author = "NVIDIA"
 
 # Single modifiable version for all of the docs - easier for future updates
-stable_version = "24.06"
-nightly_version = "24.08"
+stable_version = "24.08"
+nightly_version = "24.10"
```
jameslamb, line 25: Could we also update the `stable` section below to use CUDA 12.5 and Python 3.11?

For #377.

That's what we tested on the nightly version of the docs, in #402 .

https://github.com/rapidsai/deployment/blob/e03c1385efdc3540216e4f2d32e5c7afa085e29a/source/conf.py#L27-L42

jameslamb (commented):
I have one suggestion.

Normally just updating the RAPIDS version would be enough for a release, but this time I think we should update the Python and CUDA versions too, since we tested CUDA 12.5 + Python 3.11 in the release testing.

jameslamb:
blegh I was too late 😭 

I can push a follow-up.

PR #415, opened 2024-08-13
remove unused images

jameslamb:
I looked through the images in `source/_static/`, found a few that appear to be unused. This proposes removing them.

## Notes for Reviewers

Searched for these like this:

```shell
git grep -E 'jupyter\-lab'
git grep airline_data
git grep -E 'header\.png'
git grep sagemaker_notebook
git grep 2_over
```


jacobtomlinson (approved):
Thanks @jameslamb

PR #416, opened 2024-08-20
Add unused images pre-commit check

jacobtomlinson:
Inspired by @jameslamb's housekeeping in #415 I thought I'd add a little pre-commit hook that checks for unused images. This way if we clean up pages in the future without removing images they will get flagged when committing.

This also caught another unused image.


scripts/unused_images.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
```
jameslamb, line 2:
```suggestion

set -e -u -o pipefail

```

Would you consider adding these to ensure that errors are raised immediately and that unset variables aren't silently treated as empty strings?

jameslamb (approved):
🤩 love it! I left one small suggestion, do whatever you want with it.

Thanks for doing this!

jacobtomlinson:
Great suggestion thanks @jameslamb

PR #426, opened 2024-09-20
added NIMs section

melodywang060:
work-in-progress section for NIMs in the rapids.ai deployment webpage
went through a lot of people to confirm, but so far NeMo Retriever seems to be the only publicly released NIM that is built on RAPIDS
we can add more when they are released


source/nims/index.md
```diff
@@ -0,0 +1,22 @@
+---
+review_priority: "index"
+html_theme.sidebar_secondary.remove: true
+---
+
+# NIMs Microservices
```
jacobtomlinson, line 6: I think we should try and match the exact wording that is being used on nvidia.com. They seem to refer to them as `NVIDIA NIM Microservices`.

https://docs.nvidia.com/nim/index.html

```suggestion
# NVIDIA NIM Microservices
```

source/index.md
```diff
@@ -98,4 +98,16 @@ Detailed guides on how to deploy and optimize RAPIDS.
 {bdg}`Infiniband`
 {bdg}`MIG`
 ````
+
+````{grid-item-card}
+:link: nims/index
+:link-type: doc
+{fas}`zap;sd-text-primary` NVIDIA NIMs
```
jacobtomlinson, line 105:
```suggestion
{fas}`zap;sd-text-primary` NVIDIA NIM Microservices
```

source/index.md
```diff
@@ -98,4 +98,16 @@ Detailed guides on how to deploy and optimize RAPIDS.
...
+^^^
+NIM Microservices using RAPIDS to accelerate your AI deployment.
```
jacobtomlinson, line 107:
```suggestion
NVIDIA NIM Microservices using RAPIDS to accelerate your AI deployment.
```

source/index.md
```diff
@@ -98,4 +98,16 @@ Detailed guides on how to deploy and optimize RAPIDS.
...
+
+{bdg}`NeMo Retriever`
+{bdg}`---`
+{bdg}`---`
```
jacobtomlinson, line 111: I wonder if these labels should be categories of NIMs rather than names fospecific NIMs. For example we could add a `natural language processing` label, and a `data processing` label.

source/nims/index.md
```diff
@@ -0,0 +1,22 @@
...
+
+`````{gridtoctree} 1 2 2 3
+:gutter: 2 2 2 2
+
+````{grid-item-card}
+:link: nemo
+:link-type: doc
```
jacobtomlinson, line 13: I'm undecided whether we should be lining to a page on our docs here, or whether we should just link directly to the NIM. The page will just be repeating the information from the NIM page, so maybe a direct link would be better. What do you think?

jacobtomlinson (commented):
This looks great. I make a few comments.

Also can you look at the linting and build failures. Let me know if you need any help deciphering the error messages.

source/index.md
melodywang060: I agree - this would help users unfamiliar with NIMs services to identify which service they might need for a particular task they have. It's totally more purpose driven this way 👍🏼

source/nims/index.md
melodywang060: Just tried the alternative link in my most recent commit - I think I like it better since it's definitely more straightforward - what about you?

source/index.md
```diff
@@ -98,4 +98,16 @@ Detailed guides on how to deploy and optimize RAPIDS.
...
+````{grid-item-card}
+:link: nims/index
+:link-type: doc
+{fas}`zap;sd-text-primary` NVIDIA NIM Microservices
+^^^
+NVIDIA NIM Microservices using RAPIDS to accelerate your AI deployment.
+
+{bdg}`Natural Language Processing`
+{bdg}`Data Processing`
+{bdg}`---`
```
jacobtomlinson, line 111: Let's get rid of this placeholder line.

```suggestion
```

source/nims/nemo.md
```diff
@@ -0,0 +1,11 @@
+# NeMo Retriever
```
jacobtomlinson, line 1: I think we can just delete this file now.

jacobtomlinson (commented):
I think the direct link works well. This also means we can drop the `nemo.md` file now as we don't link to it anywhere.

This just leaves `nims/index.md` and if we aren't going to put other files in that directory we might just want to rename it to `nims.md` at the top level and avoid the directory altogether.

jacobtomlinson (approved):
This looks great, thanks!

PR #429, opened 2024-09-25
Convert tasklist to checkbox list

### LINKED ISSUE #428: Tasklists being replaced with sub-issues
Each release we open an issue that contains a tasklist of all the deployment docs pages. We work our way through the list checking off pages as we test them. See #402 as an example.

If a page isn't right we convert the task to an issue and then open a PR that closes the issue.

Image

We have a [script that generates the issue](https://github.com/rapidsai/deployment/blob/main/scripts/gen_release_checklist_issue.py) which includes the tasklists.

https://github.com/rapidsai/deployment/blob/a2e7fe669e93010ffa62553d1b4681aa67d35a63/scripts/gen_release_checklist_issue.py#L75-L83

It seems that GitHub is removing tasklists in favour of regular lists and subissues. We should be able to remove the tasklist code fence and just use a regular checkbox list from now on as they now have a menu that provides a "convert to issue" button. 

Image

jacobtomlinson:
Closes #428 

Updated the script that generates the release checklist issue. Removed the tasklist markdown formatting in favour of a regular checkbox list as tasklists are being replaced with sub-issues. Sub-issues can be created directly by converting regular checkbox list items now.

Previously a list would be generated that looks like this.

````markdown
```tasklist
### List Title
- [ ] Foo
- [ ] Bar
- [ ] Baz
```
````

But now we can simplify it. We still want multiple lists with titles, but we don't need the code fence.

````markdown
### List Title

- [ ] Foo
- [ ] Bar
- [ ] Baz
````


jameslamb (approved):
Sad, I really liked the GitHub tasklist! Oh well, if they're changing it, nothing we can do.

jacobtomlinson:
I was initially sad too, but after digging deeper it doesn't seem like we are losing any functionality.

You can still create lists with checkboxes, then convert those items to an issue in one click. And any checkbox that is a link to another issue results in the "tracked by" field being populated on that issue. 

I think the only thing that has gone is the ability to add an item to the list directly from the rendered issue. To add new items you need to edit the issue and add the item in markdown.

So it's nearly all the same things, but more deeply integrated.

jameslamb:
>  And any checkbox that is a link to another issue results in the "tracked by" field being populated on that issue

ooooo that's VERY nice! Ok you're right, I'm not sad any more 😁

PR #430, opened 2024-09-25
update build and test dependencies

jameslamb:
Since we're nearing another period of heavy activity in this project, testing the 24.10 release, this proposes updating the build and test dependencies here to their latest versions, similar to #396.

* conda env for docs building:
   - *Python `3.11 -> 3.12`*
   - *all other dependencies with floors on their latest versions (for faster conda solves)*
   - *limiting to only `conda-forge`*
* pre-commit hooks to their latest versions
  - *via `pre-commit autoupdate`
* GitHub Actions configs:
   - *Python `3.11 -> 3.12`*
   - *all third-party actions to their latest versions (no changes here, just saying I checked this)*

### DIFF
--- a/conda/environments/deployment_docs.yml
+++ b/conda/environments/deployment_docs.yml
@@ -3,16 +3,16 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - myst-nb>=1.1.1
-  - myst-parser>=3.0.1
-  - nbsphinx>=0.9.4
-  - numpydoc>=1.7.0
-  - pydata-sphinx-theme>=0.12.0
-  - python=3.11
-  - pre-commit>=3.7.1
-  - sphinx>=7.4.0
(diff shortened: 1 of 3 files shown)


conda/environments/deployment_docs.yml
```diff
@@ -1,18 +1,18 @@
 name: deployment-docs-dev
 channels:
+  - nodefaults
```
jameslamb, line 3: See https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#creating-an-environment-file-manually

> *You can exclude the default channels by adding `nodefaults` to the channels list.*

This will force the entire solve to be done with only `conda-forge`. That's generally faster than mixing multiple channels, because there's less repodata to download and fewer candidates packages to consider when trying to solve.

jacobtomlinson (approved):
Looks great thanks @jameslamb

PR #436, opened 2024-10-03
simplify markdownlint configuration

jameslamb:
#74 introduced an empty `.markdownlintignore` file, and that file appears to have been empty ever since.

I think it's safe to remove, and that doing so would slightly simplify this repo.

This proposes that removal.

## Notes for Reviewers

Let's not merge until @jacobtomlinson specifically reviews... there might be some context I'm missing that explains why this empty file was added back in #74.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,6 @@ repos:
     rev: v0.42.0
     hooks:
       - id: markdownlint
-        args: [--ignore-path=.markdownlintignore]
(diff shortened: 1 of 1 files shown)


jacobtomlinson (commented):
I think I added this just for future use. The pre-commit example probably just included the args that specified it and I added an empty file for good measure.

I have no strong feelings either way on removing it.

jameslamb:
Alright thanks! I think we should remove it then, to simplify things a bit.

jacobtomlinson (approved):

PR #451, opened 2024-10-07
Google Colab: clarify language, remove outdated links

jameslamb:
I ran through the colab demo today (https://docs.rapids.ai/deployment/nightly/platforms/colab), with the 24.10 nightlies.

Even found a bug in the installer! (https://github.com/rapidsai-community/rapidsai-csp-utils/pull/112)

This proposes the following changes based on that experience:

* remove links that are overly-general or that point to archived repos
* be consistent about use of "RAPIDS" vs. "RAPIDSAI"
* point to the install selector instead of enumerating GPU types

### DIFF
--- a/source/platforms/colab.md
+++ b/source/platforms/colab.md
@@ -53,13 +53,13 @@ Choose GPU for Hardware Accelerator
 
 ### 3. Check GPU type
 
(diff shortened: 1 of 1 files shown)


melodywang060 (approved):
good bug find!

jacobtomlinson (approved):

PR #454, opened 2024-10-08
deleted $ for multiline commands

### LINKED ISSUE #452: Remove Excess `$` From Shell Blocks
##  Problem
In a lot of code chunks in RAPIDs deployment documentation, only the first line of the multiline command is copied over when users hit the copy code block button - the rest are omitted 

## Example: 

```
$ gcloud container clusters create rapids-gpu-kubeflow \
  --accelerator type=nvidia-tesla-a100,count=2 --machine-type a2-highgpu-2g \
  --zone us-central1-c --release-channel stable
```

In the example above, only the first line is interpret as the command, and the following lines are interpreted as the console output from the command in the first line

## Quick Fix
Removing the `$` in the beginning of the code chunk 

This interprets the multiline command as a whole block of code to copy over instead of just the first line. 

## Pros
- More convenient for users to copy and paste

melodywang060:
fixes #452

### DIFF
--- a/source/cloud/azure/aks.md
+++ b/source/cloud/azure/aks.md
@@ -23,7 +23,7 @@ $ az login
 Now we can launch a GPU enabled AKS cluster. First launch an AKS cluster.
 
 ```console
-$ az aks create -g <resource group> -n rapids \
+    az aks create -g <resource group> -n rapids \
         --enable-managed-identity \
         --node-count 1 \
         --enable-addons monitoring \
@@ -92,7 +92,7 @@ $ az extension add --name aks-preview
 `````
 
 ```console
-$ az aks nodepool add \
+    az aks nodepool add \
     --resource-group <resource group> \
     --cluster-name rapids \
     --name gpunp \
--- a/source/cloud/azure/aks.md
+++ b/source/cloud/azure/aks.md
@@ -23,7 +23,7 @@ $ az login
 Now we can launch a GPU enabled AKS cluster. First launch an AKS cluster.
 
 ```bash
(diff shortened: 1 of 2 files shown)


source/cloud/azure/aks.md
```diff
@@ -23,7 +23,7 @@ $ az login
 Now we can launch a GPU enabled AKS cluster. First launch an AKS cluster.
 
 ```console
```
jacobtomlinson, line 25: As this is not a `console` style example (command with output) maybe we should tell the highlighter that it's `bash` instead.

```suggestion
```bash
```

source/cloud/azure/aks.md
```diff
@@ -23,7 +23,7 @@ $ az login
...
-$ az aks create -g <resource group> -n rapids \
+    az aks create -g <resource group> -n rapids \
```
jacobtomlinson, line 26: Let's remove these leading spaces.

```suggestion
az aks create -g <resource group> -n rapids \
```

source/cloud/azure/aks.md
```diff
@@ -92,7 +92,7 @@ $ az extension add --name aks-preview
 `````
 
 ```console
```
jacobtomlinson, line 94:
```suggestion
```bash
```

jacobtomlinson (changes requested):
Thanks for getting started here @melodywang060. A couple of comments.

jacobtomlinson (changes requested):

jacobtomlinson (approved):
I tweaked a bit of whitespace, but otherwise this looks good.

PR #456, opened 2024-10-09
Fix code typo for Coiled

jacobtomlinson:
Add a couple of small tweaks to the Coiled instructions.

### DIFF
--- a/source/platforms/coiled.md
+++ b/source/platforms/coiled.md
@@ -14,7 +14,7 @@ To get started you need to install Coiled and login.
 
 ```console
 $ conda install -c conda-forge coiled
-$ coiled setup
+$ coiled login
 ```
 
 For more information see the [Coiled Getting Started documentation](https://docs.coiled.io/user_guide/getting_started.html).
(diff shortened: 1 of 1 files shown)


melodywang060 (approved):

source/platforms/coiled.md
```diff
@@ -82,7 +82,7 @@ We can also connect a Dask client to see that information for the workers too.
 ```python
 from dask.distributed import Client
 
-client = Client
+client = Client(cluster)
```
charlesbluca, line 85: Coming around to this PR now as I'm going through the Coiled deployment process - is this example meant to be specific to the Jupyter server case, or more an illustration of how to connect a Dask client generally?

Depending on the answer, might be worth clarifying here that (the screenshot below seems to imply) we can just use `Client()` from the Jupyter server to automatically connect to the spawned cluster

jacobtomlinson: That's a good point. The correct fix here would've just been `Client()` because this was referring to using it within the Jupyter session. @charlesbluca would you mind making a follow-up PR here?

charlesbluca: Yeah sure! Fixed in https://github.com/rapidsai/deployment/pull/473

PR #457, opened 2024-10-09
Alternative Dask-CloudProvider Azure Installation Command 

melodywang060:
## Problem 

Currently when trying to install Dask-CloudProvider Azure, I am getting a `no matches found` error:
image

## Quick Fix

After some tweaking and exploration, I tried inserting quotation marks around the package name, and that worked. 

Therefore, in this PR I propose adding a sentence in the docs that refers to this potential error and suggests users to add quotation marks if they encounter it.

### DIFF
--- a/source/cloud/azure/aks.md
+++ b/source/cloud/azure/aks.md
@@ -23,7 +23,7 @@ $ az login
 Now we can launch a GPU enabled AKS cluster. First launch an AKS cluster.
 
 ```console
-$ az aks create -g <resource group> -n rapids \
+    az aks create -g <resource group> -n rapids \
         --enable-managed-identity \
         --node-count 1 \
         --enable-addons monitoring \
@@ -92,7 +92,7 @@ $ az extension add --name aks-preview
 `````
 
 ```console
-$ az aks nodepool add \
+    az aks nodepool add \
     --resource-group <resource group> \
     --cluster-name rapids \
     --name gpunp \
--- a/source/cloud/azure/azure-vm-multi.md
+++ b/source/cloud/azure/azure-vm-multi.md
@@ -12,6 +12,8 @@ Dask Cloud Provider can be installed via `conda` or `pip`. The Azure-specific ca
 $ pip install dask-cloudprovider[azure]
 ```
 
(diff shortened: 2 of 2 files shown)


source/cloud/azure/azure-vm-multi.md
```diff
@@ -12,6 +12,8 @@ Dask Cloud Provider can be installed via `conda` or `pip`. The Azure-specific ca
 $ pip install dask-cloudprovider[azure]
 ```
 
+Try running `pip install "dask-cloudprovider[azure]"` instead if you encounter a `zsh: no matches found` error.
```
jameslamb, line 15: Instead of adding another line to the docs, I think it'd be simpler to just add single quotes to the code sample above. That should always work, in any shell. Could you please do that instead?

source/cloud/azure/aks.md
```diff
@@ -92,7 +92,7 @@ $ az extension add --name aks-preview
 `````
 
 ```console
-$ az aks nodepool add \
+    az aks nodepool add \
```
jameslamb, line 95:
```suggestion
az aks nodepool add \
```

Can you please remove this leading space?

source/cloud/azure/aks.md
```diff
@@ -23,7 +23,7 @@ $ az login
 Now we can launch a GPU enabled AKS cluster. First launch an AKS cluster.
 
 ```console
-$ az aks create -g <resource group> -n rapids \
+    az aks create -g <resource group> -n rapids \
```
jameslamb, line 26:
```suggestion
az aks create -g <resource group> -n rapids \
```

jameslamb (changes requested):
Thanks for investigating this! Left some comments for your consideration.

In general, anywhere you see an installation requirement that uses shell characters, like these:

```text
pip install dask[cloudprovider]
pip install pandas==2.*
pip install numpy>=1.23,<3.0.0a0
```

They should be wrapped in single quotes to prevent the shell from assigning special meaning to those characters.

If you see other cases like this in the docs, I'd support adding single quotes to them (does not need to be in this PR though).

PR #461, opened 2024-10-10
AzureML hpo Workbook 

melodywang060:
## Problem 

Some of the configurations for the ML client was hardcoded for our team-specific subscriptions, but users will need to change that for their own dependencies. However, that isn't clearly outlined in the guide and users might end up copying and pasting the whole thing over without much thought

## Quick Fix

Omitting our team-specific configurations and replacing with #FILL IN blanks will aid users in identifying which places in the code need to be manually configured for their specific use cases.


mmccarty (approved):
Looks good. Thanks!

source/cloud/azure/aks.md
```diff
@@ -91,8 +91,8 @@ $ az extension add --name aks-preview
 
 `````
 
-```console
-$ az aks nodepool add \
+```bash
+    az aks nodepool add \
```
jameslamb, line 95:
```suggestion
az aks nodepool add \
```

jameslamb (approved):

jacobtomlinson:
I think these changes were included as part of another PR.

Issue #463, opened 2024-10-10
bug: dev conda enviornment not working with sphinx 8.1.0


ncclementi:
I was trying to install the dev dependencies following the instruction from the readme, and with the latest release (v8.1.0 released recently) I get an error when trying to do `make dirhtml`

```bash
Running Sphinx v8.1.0
loading translations [en]... done

Extension error:
Could not import extension sphinxcontrib.mermaid (exception: cannot import name 'ExtensionError' from 'sphinx.util' (.../miniforge3/envs/deployment-docs-dev/lib/python3.12/site-packages/sphinx/util/__init__.py))
make: *** [dirhtml] Error 2
```

If I downgrade to `sphinx=8.0.2` the problem is resolved. Not sure exactly what the problem is but a potential temporary solution would be to pin sphinx. Happy to open a PR, if you think this is the right way to do go about it.

jacobtomlinson:
Thanks for raising this @ncclementi! Yeah it looks like `8.1.0` released yesterday and broke the build here. Pinning would be good but I expect a fix either needs to go into `sphinx` or `sphinxcontrib-mermaid ` to fix this properly.

jacobtomlinson:
I've opened #469 to add an upper bound to Sphinx and also https://github.com/mgaitan/sphinxcontrib-mermaid/pull/161 to fix this upstream.

jameslamb:
Thanks for handling this @jacobtomlinson !

PR #466, opened 2024-10-10
Fix broken links and some typos

ncclementi:
I was going over the docs and found multiple broken links and some typos, so I went ahead and fix them. In some situations I wasn't sure about the correct link, please double check they are pointing to were they are supposed to.

### DIFF
--- a/source/tools/kubernetes/dask-operator.md
+++ b/source/tools/kubernetes/dask-operator.md
@@ -1,7 +1,7 @@
 # Dask Operator
 
 Many libraries in RAPIDS can leverage Dask to scale out computation onto multiple GPUs and multiple nodes.
-[Dask has an operator for Kubernetes](https://kubernetes.dask.org/en/latest/operator.html) which allows you to launch Dask clusters as native Kubernetes resources.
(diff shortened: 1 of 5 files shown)


source/tools/kubernetes/dask-operator.md
```diff
@@ -226,7 +226,7 @@ spec:
 ```
 
 For the scheduler pod we are also setting the `rapidsai/base` container image, mainly to ensure our Dask versions match between
-the scheduler and workers. We also disable Jupyter and ensure that the `dask-scheduler` command is configured.
+the scheduler and workers. We ensure that the `dask-scheduler` command is configured.
```
ncclementi, line 229: I think there is something missing in the yaml file that actually disables Jupyter, but it's not clear to me. Please correct if I'm wrong.

ncclementi:
The CI failure is related to https://github.com/rapidsai/deployment/issues/463. there is a problem with the new sphinx release v8.1

jacobtomlinson (approved):
Thanks Naty this looks great!

PR #469, opened 2024-10-11
Install sphinxcontrib-mermaid from proposed sphinx 8.1.0 compat fix

jacobtomlinson:
This PR tests https://github.com/mgaitan/sphinxcontrib-mermaid/pull/161 which should be a fix for #463.


jacobtomlinson:
I'm going to close this in favour of #470 as that seems like a more robust way to deal with this breakage for now. We can rever that PR once the changes in https://github.com/mgaitan/sphinxcontrib-mermaid/pull/161 have been released.

PR #472, opened 2024-10-11
fix image rendering in rapid-sagemaker-hpo notebook

### LINKED ISSUE #468: Images not rendering in rapids-sagemaker-hpo notebook
In https://docs.rapids.ai/deployment/stable/examples/rapids-sagemaker-hpo/notebook/ I don't see the images rendering (tried 3 different browser and I have the same problem)

For example:

Image

I can't reproduce locally, in the local build they do render properly. 

I think the problem is because images are inserted using this format 

```python
<img src='../../_static/images/examples/rapids-sagemaker-hpo/hpo.png'>
``` 

instead of this

```python
![](../../_static/images/examples/rapids-sagemaker-hpo/hpo.png)
```

but I'm not entirely sure, if someone can confirm they see the same and it's not just on my end, I'm happy to open a PR.

ncclementi:
Closes #468 

Let's give it a try and see what happens in the rendering when using markdown-style hyperlinks.  

Edit: I check the read the doc link and looks like it works 🚀

### DIFF
--- a/source/examples/rapids-sagemaker-hpo/notebook.ipynb
+++ b/source/examples/rapids-sagemaker-hpo/notebook.ipynb
@@ -43,7 +43,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "<img src='../../_static/images/examples/rapids-sagemaker-hpo/hpo.png'>"
+    "![](../../_static/images/examples/rapids-sagemaker-hpo/hpo.png)"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
Thanks for this! Looks good to me in the docs too.

And I checked the repo for other stuff like this, didn't find any more.

```shell
git grep -E 'img.*src'
```

Appreciate the help!

PR #474, opened 2024-10-14
Replace sphinx upper-bound with sphinx-mermaid lower-bound

jacobtomlinson:
Reverts #470 and replaced the upper-bound on `sphinx` with a lower-bound on `sphinx-mermaid`.


jameslamb (approved):
I've been following the conversation in https://github.com/mgaitan/sphinxcontrib-mermaid/issues/160 and similar, this seems like the right fix. Thanks!

PR #482, opened 2024-12-10
Tweak Databricks install instructions

jacobtomlinson:
- Made it clear you need to click "Add" for the user script
- Updated pip commands
- Added more templating for pip commands

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -33,7 +33,7 @@ versions = {
         "rapids_conda_channels": "-c rapidsai -c conda-forge -c nvidia",
         "rapids_conda_packages": f"rapids={stable_version} python=3.12 cuda-version=12.5",
         "rapids_pip_index": "https://pypi.nvidia.com",
-        "rapids_pip_version": f"{nightly_version}.0a0,<={nightly_version}",
+        "rapids_pip_version": f"{stable_version}",
     },
     "nightly": {
         "rapids_version": f"{nightly_version}-nightly",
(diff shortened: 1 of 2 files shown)


source/conf.py
```diff
@@ -40,6 +42,8 @@
         "rapids_notebooks_container": f"rapidsai/notebooks:{nightly_version + 'a'}-cuda12.5-py3.12",
         "rapids_conda_channels": "-c rapidsai-nightly -c conda-forge -c nvidia",
         "rapids_conda_packages": f"rapids={nightly_version} python=3.12 cuda-version=12.5",
+        "rapids_pip_index": "https://pypi.anaconda.org/rapidsai-wheels-nightly/simple",
+        "rapids_pip_version": f"{nightly_version}.0a0,<={nightly_version}",
```
jameslamb, line 46:
```suggestion
        "rapids_pip_version": f"{nightly_version}.*,>=0.0.0a0",
```

As currently spelled, this would always pull in the first nightly (*exactly* `24.12.0a0`).

Instead, we want this to be like `24.12.*,>=0.0.0a0`. That's the form that we use throughout RAPIDS ([example from cudf](https://github.com/rapidsai/cudf/blob/657f50bae866d97a231d565f34a1941efd49c721/dependencies.yaml#L614)), to limit to a single release and to allow pre-release versions to be found without needing to pass `--pre` to `pip install`.

jameslamb (approved):
Left one change request, committing it (to start CI) and approving, assuming you'll agree. I'll leave this to you to merge though, in case you disagree.

PR #490, opened 2025-01-09
Check pynvml works after move to nvidia-ml-py

ncclementi:
Closes https://github.com/rapidsai/deployment/issues/151

These are the places where we import, link or mention pynvml https://github.com/search?q=repo%3Arapidsai%2Fdeployment+pynvml&type=code 

As part of this PR I manually went over these with a RAPIDS 25.02 nightly installation and check they worked as expected.

Questions remaining: 
-  Here we are installing `pynvml` but we also install `rapids` which comes with `pynvml` do we want to remove then specific package or replace it for `nvidia-ml-py` https://github.com/rapidsai/deployment/blob/370c7222c0882f3fecc57f3bcc267f7cfba249ea/source/guides/azure/infiniband.md?plain=1#L261 

- Do we want to try this in azure itself, or the confirmation of running a similar installation locally and running this is sufficient? 
https://github.com/rapidsai/deployment/blob/370c7222c0882f3fecc57f3bcc267f7cfba249ea/source/cloud/azure/azure-vm-multi.md?plain=1#L57

### DIFF
--- a/source/guides/mig.md
+++ b/source/guides/mig.md
@@ -30,7 +30,7 @@ GPU 0: NVIDIA A100-PCIE-40GB (UUID: GPU-84fd49f2-48ad-50e8-9f2e-3bf0dfd47ccb)
 
 In the example case above the system has one NVIDIA A100 with 3 x 10GB MIG instances. In the next sections we will see how to use the instance names to startup a Dask cluster composed of MIG GPUs. Please note that once a GPU is partitioned, the physical GPU (named `GPU-84fd49f2-48ad-50e8-9f2e-3bf0dfd47ccb` above) is inaccessible for CUDA compute and cannot be used as part of a Dask cluster.
 
-Alternatively, MIG instance names can be obtained programmatically using [NVML](https://developer.nvidia.com/nvidia-management-library-nvml) or [PyNVML](https://github.com/gpuopenanalytics/pynvml). Please refer to the [NVML API](https://docs.nvidia.com/deploy/nvml-api/) to write appropriate utilities for that purpose.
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
@jakirkham might be able to answer the question about the metapackage. Do we need to explicitly install this at all?

 I don't think we need to try things on Azure directly, although you can if you want to. The main thing is checking the code snippet works on a machine with `nvidia-ml-py` installed.

ncclementi:
I tried to check this on a VM in particular an Azure VM, and I noticed that when installing rapids-notebook nightly via docker.

```bash
docker run --gpus all --pull always --rm -it \
    --shm-size=1g --ulimit memlock=-1 --ulimit stack=67108864 \
    -p 8888:8888 -p 8787:8787 -p 8786:8786 \
    rapidsai/notebooks:25.02a-cuda12.5-py3.12
```

It comes with an older `pynvml`, meaining there is still some libraries that might be pinned with `<12.0` still. 

In the VM via docker, we get 

```python
>>> import pynvml
>>> pynvml.__version__
'11.5.3'
``` 

That being said, I upgraded it once in the VM and `pynvml` worked as expected.

jakirkham:
It looks like the Docker nightly image builds have been failing. Appears there is some issue downloading datasets for the cuVS benchmark step. Have filed an upstream issue: https://github.com/rapidsai/docker/issues/724

So think this just comes down the Docker images being a bit stale as it has been a bit since we have had a successful nightly build there

Will take a note to mention this to the OPS and cuVS teams on Monday. Maybe someone can provide some guidance on how we can address this to get the Docker builds up and running again

jacobtomlinson (approved):
Thanks for testing @ncclementi. I think given that you upgraded it and things seemed ok I'm happy to merge this.

jakirkham:
Thanks Naty and Jacob! 🙏

Issue #419, opened 2024-09-05
Document running the RAPIDS Notebooks container as a Snowpark Service


jacobtomlinson:
The Snowflake documentation has an [example of running a container with Jupyter in as a service in Snowpark](https://quickstarts.snowflake.com/guide/intro_to_snowpark_container_services/ ) and accessing the Jupyter UI via a Service endpoint.

It would be great to add a documentation page that walks through the same steps but using the RAPIDS notebooks container on a GPU compute pool.

I imagine the steps will be something along the lines of:

- Build an image based on the RAPIDS notebooks image and install the [snowflake Python library](https://docs.snowflake.com/en/developer-guide/python-connector/python-connector)
- Push the image to the Snowflake container registry
- Create a service specification with the RAPIDS container
- Launch the service
- Expose/access the Jupyter UI
- Show reading data from Snowflake into a cudf dataframe

ncclementi:
I looked a bit more into this, and it looks like the steps are pretty much what's detailed. Couple of notes and things to figure out are: 

Here is a walkthourgh video that also covers part of the rpocess: https://youtu.be/izeG3pvvy9M?list=PLavJpcg8cl1GN7AwaY95PXqwUw4qj8zLo&t=411
> Snowflake setup: 

TODO: 
- [ ] Access to non-trial Snowflake account
It looks like to do most of this, it's needed A Snowflake account login with a role that has the ACCOUNTADMIN and a certain type of
- [ ] Setup snowflake env https://quickstarts.snowflake.com/guide/intro_to_snowpark_container_services/#1

> local env setup/config: 
- configure snowflake cli and create a connection to the container
- start docker and check login into image repo registry
ref: https://quickstarts.snowflake.com/guide/intro_to_snowpark_container_services/#2

> Build an image based on the RAPIDS notebooks image and install the [snowflake Python library](https://docs.snowflake.com/en/developer-guide/python-connector/python-connector)

TODO: 
- [ ] Determine what extra dependencies need to be added  (probably "snowflake-snowpark-python[pandas]" snowflake-connector-python ? )
- [ ] Create docker file that is based on the RAPIDS notebook image, create the image

 For what I could gather from the demos, it seems like you build the image locally and then push it to the container registry. I was hoping one would be able to pull from let's say dockerhub or similar directly into snowpark registry but I can't find docs on that. Which it might be a bit limiting for a doc example  

> Push the image to the Snowflake container registry

[...]

PR #493, opened 2025-01-16
Rapids on snowflake deployment

ncclementi:
Closes the deployment part of https://github.com/rapidsai/deployment/issues/419

For the "Show reading data from Snowflake into a cudf dataframe" , I created this issue for tracking purposes https://github.com/rapidsai/deployment/issues/494

~~This is WIP but it's almost there for review, need a few more sections with screenshots.~~

TODO: 
- [x] Figure out failures in CI. Not sure why I'm getting the failures, I can build locally although I get a warning that says< I assume we fail on warning sin CI. It seems it doesn't like the `$`. Any suggestions @jacobtomlinson ?


ncclementi:
Can't seem to find a fix for the failure: 

Looks like it is sort of a known issue  https://github.com/sphinx-doc/sphinx/issues/3175 but the suggested fix, I found relies on this sphinx syntax and when trying this, it doesn't render as expected. Any ideas?

.. code-block:: sql :force: 
   USE ROLE CONTAINER_USER_ROLE;
   CALL SYSTEM$REGISTRY_LIST_IMAGES('/CONTAINER_HOL_DB/PUBLIC/IMAGE_REPO');

source/platforms/snowflake.md
```diff
@@ -0,0 +1,303 @@
...
+- Access to `INSTANCE_FAMILY` with NVIDIA GPUs. For this guide we will use `GPU_NV_S`
+  (1 NVIDIA A10G - smallest NVIDIA GPU size available for Snowpark Containers to
+  get started.)
+
+## Set up the Snowflake environment
+
+In a SQL worksheet in Snowflake, run the following commands to create the role,
+database, warehouse, and stage that we need to get started:
+
+```sql
```
jacobtomlinson, line 22: I wonder if making this a myst directive like [this example](https://myst-parser.readthedocs.io/en/v0.17.1/syntax/syntax.html#parameterizing-directives) would help?

````suggestion
```{code-block} sql
---
force
---
````

Comment (ncclementi) [reply source/platforms/snowflake.md:22]: No but close, thanks for the idea this worked:

```{code-block} sql
:force:

CALL SYSTEM$GET_SERVICE_STATUS('CONTAINER_HOL_DB.PUBLIC.rapids_snowpark_service');
```

ncclementi:
While checking if the volume mount instructions where working correctly I realized that the mount won't work unless we have `workspace` as a  directory created. 

We have two options here. 
1. Just mount `/home/rapids` so all the changes made to the notebooks in there would remain, as well as things added. 
2. Modify the `Dockerfile`such that we create this `workspace.`  I think it would be something like this? 
Any strong preferences here?

```Dockerfile
FROM rapidsai/notebooks:25.02a-cuda11.8-py3.11-amd64

RUN pip install "snowflake-snowpark-python[pandas]" snowflake-connector-python

USER rapids
WORKDIR /home/rapids/workspace
```

source/platforms/snowflake.md
```diff
@@ -0,0 +1,343 @@
+# Snowflake
+
+You can install RAPIDS on Snowflake via [Snowpark Container Services](https://docs.snowflake.com/en/developer-guide/snowpark-container-services/overview).
```
jacobtomlinson, line 3: First mention of Snowflake should be a link.

```suggestion
You can install RAPIDS on [Snowflake](https://www.snowflake.com) via [Snowpark Container Services](https://docs.snowflake.com/en/developer-guide/snowpark-container-services/overview).
```

source/platforms/snowflake.md
```diff
@@ -0,0 +1,343 @@
...
+
+### Build Docker image locally
+
+For this guide, we build an image that starts from the RAPIDS notebook image and
+adds some extra snowflake packages.
+
+Create a Dockerfile as follow:
+
+```Dockerfile
+FROM rapidsai/notebooks:25.02a-cuda11.8-py3.11-amd64
```
jacobtomlinson, line 97: This will go stale. We usually use [templating](https://github.com/rapidsai/deployment?tab=readme-ov-file#releasing) to ensure it always shows the correct RAPIDS image. The requirements to use Python 3.11 may add a little complexity here.

source/platforms/snowflake.md
```diff
@@ -0,0 +1,343 @@
...
+```yaml
+name: snow-cli
+channels:
+  - condaforge
+dependencies:
+  - python=3.11
+  - pip
+  - pip:
+      - snowflake-cli
+```
```
jacobtomlinson, line 127: Creating a whole conda environment to install the snowflake CLI seems specific to a subset of users.

I would suggest just directing folks to the [snowflake CLI install docs](https://docs.snowflake.com/en/developer-guide/snowflake-cli/installation/installation) and letting them install it in whatever way makes sense to them.

For example my personal preference would be to use homebrew on my mac, which their docs include. Or maybe use `uv tool` to install the Python package.

source/platforms/snowflake.md
```diff
@@ -0,0 +1,343 @@
...
+
+```{note}
+If you don't recall `<ORG>-<ACCOUNT-NAME>` you can obtain them
+by running the following in the Snowflake SQL worksheet.
+```
+
+```sql
+SELECT CURRENT_ORGANIZATION_NAME(); --org
+SELECT CURRENT_ACCOUNT_NAME();      --account name
+```
```
jacobtomlinson, line 163: I didn't see this until after I spent a while trying to figure out where to find this info. Maybe move this up, or mention it in the comment next to `account : <ORG>-<ACCOUNT-NAME> # e.g. MYORGANIZATION-MYACCOUNT`

source/platforms/snowflake.md
```diff
@@ -0,0 +1,343 @@
...
+
+Test the connection:
+
+```bash
+snow connection test --connection "CONTAINER_HOL"
```
jacobtomlinson, line 168: When I ran this I got a 2-factor prompt on my phone, we might want to mention this. 

But then after approving I got an error.

```console
$ snow connection test --connection "CONTAINER_HOL"
Could not use database "CONTAINER_HOL_DB". Object does not exist, or operation cannot be performed.
```

source/platforms/snowflake.md
```diff
@@ -0,0 +1,343 @@
...
+
+## Set up the Snowflake environment
+
+In a SQL worksheet in Snowflake, run the following commands to create the role,
+database, warehouse, and stage that we need to get started:
+
+```sql
+-- Create an CONTAINER_USER_ROLE with required privileges
+USE ROLE ACCOUNTADMIN;
+CREATE ROLE CONTAINER_USER_ROLE;
```
jacobtomlinson, line 27: Should be this `CREATE OR REPLACE` to make it idempotent?

```suggestion
CREATE OR REPLACE ROLE CONTAINER_USER_ROLE;
```

source/platforms/snowflake.md
```diff
@@ -0,0 +1,343 @@
...
+GRANT MONITOR USAGE ON ACCOUNT TO  ROLE  CONTAINER_USER_ROLE;
+GRANT BIND SERVICE ENDPOINT ON ACCOUNT TO ROLE CONTAINER_USER_ROLE;
+GRANT IMPORTED PRIVILEGES ON DATABASE snowflake TO ROLE CONTAINER_USER_ROLE;
+
+-- Grant CONTAINER_USER_ROLE to ACCOUNTADMIN
+grant role CONTAINER_USER_ROLE to role ACCOUNTADMIN;
+
+-- Create Database, Warehouse, and Image spec stage
+USE ROLE CONTAINER_USER_ROLE;
+CREATE OR REPLACE DATABASE CONTAINER_HOL_DB;
```
jacobtomlinson, line 41: This line fails for me with `Insufficient privileges to operate on database 'CONTAINER_HOL_DB'`

jacobtomlinson (commented):
I struggled with some permissions error when following these instructions. Could you take a look and see wha I'm doing wrong?

Also to answer your question I wouldn't set the working directory I would just create an empty directory.

```dockerfile
RUN mkdir -p /home/rapids/workspace
```

source/platforms/snowflake.md
ncclementi: Weird, I didn't get this prompting, I guess I didn't set 2FA as part of the process? It's weird that you are running into all of this because I haven't I wonder if there is something causing the problem in our account.

source/platforms/snowflake.md
ncclementi: Well, if you are in an account sharing resources, this might overwrite a role that someone else created. I wonder if all the issues you are seeing are related to that.

source/platforms/snowflake.md
ncclementi: Yes, the requirement of python 3.11 and explicit amd64 platform will make the templating a bit more complicated. I'll take a look into it

source/platforms/snowflake.md
```diff
@@ -0,0 +1,343 @@
...
+Locally, create the following file `rapids-snowpark.yaml`:
+
+```yaml
+spec:
+  containers:
+    - name: rapids-nb-snowpark
+      image: <org-account>.registry.snowflakecomputing.com/container_hol_db/public/image_repo/rapids-nb-snowflake:dev
+      volumeMounts:
+        - name: rapids-notebooks
+          mountPath: home/rapids
```
jacobtomlinson, line 234: This should be

```suggestion
          mountPath: /home/rapids/notebooks/workspace
```

jacobtomlinson: I set the workspace to this and it worked.

image

However if I try and create a new notebook in that directory it gives me a permissions error.

image

I think this is because the UID and GID needs changing. The `rapids` user's UID is `1001`.

```console
$ id
uid=1001(rapids) gid=1000(conda) groups=1000(conda)
```

source/platforms/snowflake.md
jacobtomlinson: Note to say we figured it out and changing the names or deleting the existing resources seemed to resolve the problem.

source/platforms/snowflake.md
```diff
@@ -0,0 +1,343 @@
...
+USE ROLE CONTAINER_USER_ROLE;
+SHOW IMAGE REPOSITORIES IN SCHEMA CONTAINER_HOL_DB.PUBLIC;
+```
+
+You will see that the repository url is `org-account.registry.snowflakecomputing.com/container_hol_db/public/image_repo` where `org-account` refers to your organization and account, the `SNOWFLAKE_REGISTRY_HOSTNAME` is the url up to the `.com`. i.e. `org-account.registry.snowflakecomputing.com`
+
+First we login into the snowflake repository with docker, via terminal:
+
+```bash
+docker login <snowflake_registry_hostname> -u <snowflake_user_name>
```
jacobtomlinson, line 184: Both the `docker login` and `docker push` commands don't work if you have 2FA enabled. Every API call results in a push notification to approve, but if you don't approve within a few seconds it sends another push and the original API call times out. Then your device just gets spammed with retry notifications but the `docker login` command has already exited.

I managed to log in by hitting approve on the very first push notification, but then for the `docker push` you get a notification for every single layer push and it's imporrible to approve the first notification for each one.

Turning on [token caching](https://docs.snowflake.com/en/user-guide/security-mfa#using-mfa-token-caching-to-minimize-the-number-of-prompts-during-authentication-optional) may help here as it should only prompt until the first approval, but I'm not sure it will help with the race condition and retries.

We should probably add a note to suggest enabling token caching or disabling 2FA altogether. We should also try and report this to Snowflake.

source/platforms/snowflake.md
```diff
@@ -0,0 +1,343 @@
...
+
+If you want to cleanup completely and remove all of the objects created, run the
+following:
+
+```sql
+USE ROLE CONTAINER_USER_ROLE;
+ALTER COMPUTE POOL CONTAINER_HOL_POOL STOP ALL;
+ALTER COMPUTE POOL CONTAINER_HOL_POOL SUSPEND;
+
+DROP SERVICE CONTAINER_HOL_DB.PUBLIC.RAPIDS_SNOWPARK_SERVICE;
```
jacobtomlinson, line 335: This line didn't work for me, it just said the service didn't exist, but the rest of the commands did so I think I cleaned up successfully.

jacobtomlinson (commented):
I managed to test the instructions end-to-end and I managed to get a working notebook instance and ran some `cudf` code 🎉.

I found a few more things that need tweaking though and left a few more comments.

source/platforms/snowflake.md
ncclementi: Interesting, this is as suggested by snowflake. Did your service had a different name or something? I'll give this another try after again soon and check.

ncclementi:
I had a good amount of trouble launching things again, I think there was some cleanup and role mix up instructions that didn't remove everything. But hopefully that's sorted. 

Yet TODO:
- [ ] resolve templating of rapids image in the doc to point to latest notebook but on py 3.11 and amd platform
- [ ] Try to add a file to the workspace, I modify the spec where I change the `uid: 1001`. 
- [ ] add a paragraph about the volume mount.

ncclementi:
This should be ready now for final review, no TODOs left. Looks like I mess up something when rebasing, I think I fix it and it should be fine. 

Do we have squashing here when merging or do I need to cleanup the commits?

source/platforms/snowflake.md
```diff
@@ -0,0 +1,343 @@
...
+      protocol: TCP
+      public: true
+    - name: dask-dashboard
+      port: 8787
+      public: true
+  volumes:
+    - name: rapids-notebooks
+      source: "@volumes/rapids-notebooks"
+      uid: 1001 # rapids user's UID
+      gid: 1000
```
jacobtomlinson, line 252: I tested this persistence out by creating a notebook and deleting and recreating the service and it worked great 🎉 

image

source/platforms/snowflake.md
jacobtomlinson: I am still seeing this

image

jacobtomlinson (commented):
I ran through end-to-end and everything works great for me. I added a couple of comments but generally this is great.

source/platforms/snowflake.md
ncclementi: Great! Yes I tested it too I should have mentioned that : )

source/platforms/snowflake.md
```diff
@@ -0,0 +1,343 @@
...
+First we login into the snowflake repository with docker, via terminal:
+
+```{note}
+If you have **2FA** both `docker login` and `docker push` commands don't
+work as expected. Every API call results in a push notification to approve, but
+if not done promptly teh original API times out, complicating the completion of
+this step.
+
+We recommend enabling [token caching](https://docs.snowflake.com/en/user-guide/security-mfa#using-mfa-token-caching-to-minimize-the-number-of-prompts-during-authentication-optional)
+or disabling 2FA during this step.
```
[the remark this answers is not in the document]
ncclementi: This is a bummer, I wonder how can we report this. I'll add a line to include the 30 min hack for now.

jacobtomlinson (approved):
This looks great. I think we are at a point where we can merge this and follow up in additional PRs.

PR #498, opened 2025-01-23
Update Coiled docs to include notebooks, CLI jobs and more details on software environments

### LINKED ISSUE #186: Expand Coiled deployment documentation
I was chatting with @jacobtomlinson and @mmccarty today and Jacob mentioned that the [current Coiled deployment docs here](https://github.com/rapidsai/deployment/blob/5ebd97cb99534b4d6708b2307bdb97652fdd3150/source/platforms/coiled.md) are pretty minimal and that it would be nice to have a bit more content here. @jacobtomlinson what do you think makes sense to include here? 

cc @ntabris, from the Coiled side, for what you think might make sense to include here too 

Also, just an FYI @ntabris has been working to make gpu-enabled software environments a better experience on Coiled and I think there are also plans to update the gpu documentation in https://docs.coiled.io

### LINKED ISSUE #250: Document Coiled Notebooks and CLI Jobs
Coiled have some new features beyond Dask clusters including [notebooks](https://docs.coiled.io/user_guide/notebooks.html) and [CLI Jobs](https://docs.coiled.io/user_guide/cli-jobs.html).

We should update our documentation with examples for each of these.

jacobtomlinson:
Closes #250 
Closes #186


source/platforms/coiled.md
```diff
@@ -1,65 +1,142 @@
 # Coiled
 
-You can deploy RAPIDS on a multi-node Dask cluster with GPUs using [Coiled](https://www.coiled.io/).
+You can deploy RAPIDS on cloud VMs with GPUs using [Coiled](https://www.coiled.io/). Coiled is a software platform that manages Cloud VMs on your behalf. It manages software environments and can launch Python scripts, Jupyter Notebook servers, Dask clusters or even just individual Python functions. Remote machines are booted just in time and shut down when not in use or idle.
```
jameslamb, line 3:
```suggestion
You can deploy RAPIDS on cloud VMs with GPUs using [Coiled](https://www.coiled.io/).
Coiled is a software platform that manages Cloud VMs on your behalf.
It manages software environments and can launch Python scripts, Jupyter Notebook servers, Dask clusters or even just individual Python functions.
Remote machines are booted just in time and shut down when not in use or idle.
```

Very minor thing... would you consider splitting sentences onto their own lines? That doesn't affect how the text is rendered (markdown renderers will all still smoosh these back into one paragraph with wrapping), but it's nice for development... easier to review git diff, easier to edit the file in a text editor, and makes the git blame more useful as a source of "when did this line last change and why?" information.

source/platforms/coiled.md
```diff
@@ -1,65 +1,142 @@
...
+name: rapidsai-notebooks
+channels:
+  - rapidsai
+  - conda-forge
+  - nvidia
+dependencies:
+  # RAPIDS packages
+  - rapids={{ rapids_version }}
+  - python=3.12
+  - cuda-version>=12.0,<=12.5
```
jameslamb, line 70:
```suggestion
```

`rapids` already has a `run:` dependency on compatible versions of `cuda-version`: https://github.com/rapidsai/integration/blob/172ef624ea50670969e1fd79930a46eabdd9c3c9/conda/recipes/rapids/meta.yaml#L31

And which packages you get should happen nicely and automatically based on the detected CUDA version on the system.

I think we could probably safely remove this. That'd also mean one less thing that's likely to become out of date over time, as RAPIDS support matrix of CUDA versions shifts.

source/platforms/coiled.md
```diff
@@ -1,65 +1,142 @@
...
+```
+
+We can use this to run GPU code on a remote environment using the RAPIDS container. You can set the coiled CLI to keep the VM around for a few minutes after execution is complete just in case you want to run it again and reuse the same hardware.
+
+```concole
+$ coiled run --gpu --name rapids-demo --keepalive 5m --container {{ rapids_container }} -- python my_code.py
+...
+```
+
+This works very nicely when paired with the cudf.pandas CLI tool. For example we can run `python -m  cudf.pandas my_script` to GPU accelerate our Pandas code without having to rewrite anything. For example [this script](https://gist.github.com/jacobtomlinson/2481ecf2e1d2787ae2864a6712eef97b#file-cudf_pandas_coiled_demo-py) processes some open NYC parking data. With `pandas` it takes around a minute, but with `cudf.pandas` it only takes a few seconds.
```
jameslamb, line 102:
```suggestion
This works very nicely when paired with the cudf.pandas CLI tool. For example we can run `python -m cudf.pandas my_script` to GPU accelerate our Pandas code without having to rewrite anything. For example [this script](https://gist.github.com/jacobtomlinson/2481ecf2e1d2787ae2864a6712eef97b#file-cudf_pandas_coiled_demo-py) processes some open NYC parking data. With `pandas` it takes around a minute, but with `cudf.pandas` it only takes a few seconds.
```

jameslamb (approved):
Left some small suggestions, do what you want with them. Otherwise seems good to me (though I did not test this)

source/platforms/coiled.md
```diff
@@ -1,65 +1,142 @@
...
-import coiled
+![Screenshot of Jupyterlab running on Coiled executing some cudf GPU code](../_static/images/platforms/coiled/coiled-jupyter.png)
 
-coiled.create_software_environment(
-    name="rapids-{{ rapids_version.replace(".", "-") }}",
-    container="{{ rapids_container }}",
-)
+## Software Environments
+
+By default when running remote operations Coiled will [attempt to create a copy of your local software environment](https://docs.coiled.io/user_guide/software/sync.html) which can be loaded onto the remote VMs. While this is an excellent feature it's likely that you do not have all of the GPU software libraries you wish to use installed locally. In this case we need to tell Coiled which software environment to use.
```
ncclementi, line 37: Would launching via the notebooks container also imply that you don't need a GPU locally neither?  If so add a note about that can be helpful.

source/platforms/coiled.md
```diff
@@ -1,65 +1,142 @@
...
+### Container images
+
+All Coiled commands can be passed a container image to use. This container will be pulled onto the remote VM at launch time.
+
+```bash
+coiled notebook start --gpu --container {{ rapids_notebooks_container }}
 ```
 
-Or you can create it with a list of conda packages in case you want to customize the environment further.
+This is often the most convenient way to try out existing software environments, but is often not the most performant due to the way container images are unpacked.
```
ncclementi, line 47: This is a fair point, but if this is the way we recommend people using it, maybe we don't want to be negative about it.

ncclementi (approved):
I left couple of comments. 

The only one thing I'm a bit torn is diving into all the possible ways of creating a software environment. 

Is the goal of the doc, showcasing all the possible ways to run rapids in Coiled, or to guide people on how to get rapids up and running in the way we consider most "adequate". 

I f we want to show all the options, I'd suggest we do it such that anyone that reads this, knows what to pick. For example

Do you have a local GPU? Yes, No and based on that suggest what's their best option. 
Do you want/need a jupyter instance? Yes/No ... 

The way I read it now, I'm not sure what a user would know what to pick out of all the options. Does that makes sense?

source/platforms/coiled.md
jacobtomlinson: This environment file gets built on a build server which AFAIK doesn't have GPUs. They just set the cuda package, but it's not clear what they set it to. So if we omit this constraint I'm not sure what will happen. What do you think?

```console
$ coiled env create --help
...
  --gpu-enabled                   Set CUDA virtual package for Conda
...
```

source/platforms/coiled.md
jacobtomlinson: My intent was to document that RAPIDS should either use the container, or a prebuilt software environment. It should not use package sync.

I keps the container in most of the examples because it feels simpler and is very copy/paste friendly. But the prebuilt software environment is more performant, and it probably my recommendation as a best practice.

What do you think we should do to make this more clear?

jacobtomlinson:
Thanks for the reviews!

> Is the goal of the doc, showcasing all the possible ways to run rapids in Coiled, or to guide people on how to get rapids up and running in the way we consider most "adequate".

I think the goal is to show that RAPIDS can be used with CLI scripts, notebooks and Dask clusters. A demonstration that it is possible, rather than a demonstration of the best way to do things.

> The way I read it now, I'm not sure what a user would know what to pick out of all the options. Does that makes sense?

This is useful feedback. I think a little more about restructuring this.

source/platforms/coiled.md
ncclementi: I agree that package sync is more performant, but that implies that you have
1. a properly set environment, which is not hard
2. a GPU, if you don't package sync will not work

My guess is that someone trying to use coiled, probably doesn't have a local GPU in their laptop. I might be wrong, though. 

What we could do, is mentioned that to use package sync you need a local GPU and a properly set environment.

ncclementi: Oh I see, I remember this now. Thanks for the clarification.

Issue #382, opened 2024-06-05
remove references to outdated RAPIDS images


jameslamb:
## Description

Since https://github.com/rapidsai/docker/issues/539, the main RAPIDS docker images intended for end-user deployments have been:

* `rapidsai/base` ([DockerHub link](https://hub.docker.com/r/rapidsai/base))
* `rapidsai/notebooks` ([DockerHub link](https://hub.docker.com/r/rapidsai/notebooks))

Any references in the deployment docs to older, no-longer-supported images should be replaced.

## Benefits of this work

* reduces friction in updating to the latest RAPIDS releases
* reduces the risk of time wasted by RAPIDS users on things that are no longer supported

## Acceptance Criteria

* there are 0 remaining references to any of the images listed under "Deprecated docker repositories" in https://github.com/rapidsai/docker/issues/539
  - including in static pictures
* there are 0 remaining references to outdated images *tags* (except where keeping an old reference is an intentional choice)

## Approach

Start like this:

```shell
git grep -E 'rapidsai/'
git grep -E ':2[0-9]+\.'
```

Replace those references, wherever appropriate, with `rapidsai/base` and `rapidsai/notebooks`.

Try to do so with the templating driven by this:

https://github.com/rapidsai/deployment/blob/8bd5b05bdcfd23032daf7ee83724681416d6341d/source/conf.py#L27-L42

## Uses

List of references that need to be updated:

`rapidsai/rapidsai-core` image

* [x] [examples / rapids-autoscaling-multi-tenant-kubernetes / notebook.ipynb](https://github.com/rapidsai/deployment/blob/3a089c59043af2de207b4323cd0882d29959ce08/source/examples/rapids-autoscaling-multi-tenant-kubernetes/notebook.ipynb) (https://github.com/rapidsai/deployment/pull/501)

`rapidsai/rapidsai` image

* [x] [examples / rapids-azureml-hpo / Dockerfile](https://github.com/rapidsai/deployment/blob/3a089c59043af2de207b4323cd0882d29959ce08/source/examples/rapids-azureml-hpo/Dockerfile) (https://github.com/rapidsai/deployment/pull/484)
* [x] [examples / rapids-azureml-hpo / notebook.ipynb](https://github.com/rapidsai/deployment/blob/3a089c59043af2de207b4323cd0882d29959ce08/source/examples/rapids-azureml-hpo/Dockerfile) (https://github.com/rapidsai/deployment/pull/484)
[...]

PR #500, opened 2025-01-24
simplify and update XGBoost benchmarking example

jameslamb:
Contributes to:

* #377
* #382

Updates the XGBoost CPU vs. GPU benchmarking example to use RAPIDS 25.02 and Python 3.12.

### How I tested this

Ran this end-to-end successfully, on an x86_64 machine with CUDA 12.5 and some V100s.
Not in any particular public cloud, just a workstation I have access to.


source/examples/xgboost-rf-gpu-cpu-benchmark/notebook.ipynb
```diff
@@ -206,44 +206,23 @@
    "id": "a89edfea-ca14-4d26-94c6-0ef8eaf02d77",
```
jameslamb, line 206: <p>This change completely removes the need to mess with a <code>Dockerfile</code> or doing a <code>docker build</code> and all the complexity that comes with that.</p><p> </p><p>This example just needs an image with the RAPIDS libraries, XGBoost, and Optuna... exactly the case that the <code>EXTRA_CONDA_PACKAGES</code> support in these images was designed for! Docs on that: <a href="https://github.com/rapidsai/docker?tab=readme-ov-file#environment-variables" target="_blank">https://github.com/rapidsai/docker?tab=readme-ov-file#environment-variables</a></p><p> </p><p>I tested this today on an x86_64 machine with CUDA 12.5, found everything worked great (with the modifications in this PR).   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/deployment/pull/500/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

jameslamb: <p>Adding <code>optuna-integration</code> is also a requirement of newer versions of <code>optuna</code> (meaning this notebook has probably been broken for a bit).</p><p> </p><p>Without it, you'll see this:  </p><p> </p><blockquote>ModuleNotFoundError: Optuna's integration modules for third-party libraries have started migrating from Optuna itself to a package called <code>optuna-integration</code>. The module you are trying to use has already been migrated to <code>optuna-integration</code>. Please install the package by running <code>pip install optuna-integration</code>.  </blockquote>

jacobtomlinson (approved):
Thanks @jameslamb. I'm a little on the fence about installing at runtime, but I agree it simplifies things nicely. Let's go with this, thanks for testing it out.

jameslamb:
> I'm a little on the fence about installing at runtime,

Yeah I was too, but I think the simplification is worth it. Especially for this workload (hyper-parameter tuning), which shouldn't be too sensitive to on the order of 5-30 seconds of extra startup time.

PR #479, opened 2024-11-22
Added Feedback Link to Sidebar

melodywang060:
image

Users have the option to click on the link (which leads to the new issue page in the github repository) if they spot a mistake on the docs

PR #501, opened 2025-01-24
fix outdated RAPIDS versions

jameslamb:
For a while now, we've had issues documenting the need to updated outdated references:

* https://github.com/rapidsai/deployment/issues/382
* https://github.com/rapidsai/deployment/issues/377

These have been open so long because I've been approaching them like "each example we update should be tested end-to-end", and I personally have found it difficult to find the time to do that very time-intensive testing.

In this PR, I'm proposing that we just update all the remaining references **without manually testing them**.  Along the way here, I've also made changes to take better advantage of the automatic templating in this repo so that these are kept up to date with future RAPIDS releases.

## Notes for Reviewers

**Benefits of this:**

* avoids wasting users' time trying examples on very-old RAPIDS versions
* nudges users toward newer versions of RAPIDS, CUDA, and Python (which should generally be better)

**Why this is probably ok to do:**

* we have an easy mechanism for submitting feedback (#479)
* it's unlikely that here in 2025, some of these examples using RAPIDS 23.x still work without any modification... these cloud services tend to change frequently!

**How to Review This**

To be absolutely clear: **I did not test these changes**. Opening this to get your opinions (@ncclementi @jacobtomlinson) on just updating these things anyway, and relying on later manual testing or user feedback to uncover issues.


jacobtomlinson (approved):
This is fine with me, thanks @jameslamb. Like you say I think the feedback mechanism makes this less of an issue.

PR #511, opened 2025-02-06
Update to newer Databricks ML runtime

### LINKED ISSUE #513: Databricks ML Runtime 15 fails to import cudf
Reproducer:
- Follow the [docs here](https://docs.rapids.ai/deployment/nightly/platforms/databricks/#single-node-gpu-notebook-environment)
  - Create a startup script (see below)
  - Create single node compute instance
  - Choose the `15.4 LTS ML (includes Apache Spark 3.5.0, GPU, Scala 2.12)`
  - Choose a GPU VM type
  - Under **advanced** add the startup script

```bash
#!/bin/bash
set -e

# Install RAPIDS libraries
pip install \
    --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple \
    "cudf-cu11>=25.02.*,>=0.0.0a0" "cuml-cu11>=25.02.*,>=0.0.0a0" \
    "dask-cuda>=25.02.*,>=0.0.0a0"
```

Importing `cudf` fails

Image

jacobtomlinson:
The documentation lists the 14.2 ML Runtime. This PR bumps to the newer 15.4 ML Runtime and updates packages to CUDA 12.

I also removed a screenshot that had the runtime version in it. We have to update this screenshot every time we use a newer version, and the screenshot didn't add a huge amount of value. So it reduces maintenance effort to just remove it.

Closes #513


jacobtomlinson:
There is also a 15 LTS runtime (and 16 LTS Beta runtime), we might want to try those out too. Marking as draft until I've tested those. IIRC the 15 Beta didn't work when I last tested it.

jacobtomlinson:
I tried 15 and it failed. I'll open a new issue to track that. I think we should still merge this as is though.

jacobtomlinson:
I tried 15 again with CUDA 12 packages and things are working. I've updated the PR with these newer versions.

ncclementi (approved):
Just to double check, what didn't work before was the 15 Photon version with cuda 11? 
But 15 LTS with cuda 12 is good.

jacobtomlinson:
Not Photon, that should always be disabled. This is the Databricks ML Runtime, which is effectively an opinionated Python environment.

The 14 Runtime worked with CUDA 11, but the 15 runtime has switched to CUDA 12.

Happy to merge with that clarification?

ncclementi:
> The v14 Runtime worked with CUDA 11, but the v15 runtime has switched to CUDA 12.
> Happy to merge with that clarification?

Sounds good, I think the more info we can provide for folks to not struggle the better making choices the better.

Issue #510, opened 2025-02-06
Fix redundant tags in workflows


ncclementi:
We have some redundant things in the page

**Data Storage and Storage**

![Image](https://github.com/user-attachments/assets/71b90664-e05f-4c64-85a6-bcc2a2665f32)

![Image](https://github.com/user-attachments/assets/7e129fe6-3b32-484b-8754-e75ad1ad6001)

**Platform and Platforms**

![Image](https://github.com/user-attachments/assets/bbb9c846-7624-40e4-9faf-03dbd79a1d57)

**and Workflow and Workflows**

![Image](https://github.com/user-attachments/assets/ec6338c8-e6a7-4f44-bfd7-27716492512e)

PR #512, opened 2025-02-06
Fix redundant tagging in workflow page

ncclementi:
Closes https://github.com/rapidsai/deployment/issues/510


jacobtomlinson (commented):
Looks great!

I think we also need to prefix the CSP tags with `cloud/`.

https://github.com/rapidsai/deployment/blob/b520d417d8680564568e66010b072d1fe78d79e1/source/examples/rapids-1brc-single-node/notebook.ipynb?short_path=bef3729#L9-L14

ncclementi:
Ready for final review.

PR #484, opened 2024-12-10
Make AzureML examples more self-contained

jameslamb:
Contributes to:

* #480 
* #382
* #377

I worked through these AzureML examples:

* https://docs.rapids.ai/deployment/nightly/cloud/azure/azureml/
* https://docs.rapids.ai/deployment/nightly/examples/rapids-azureml-hpo/notebook/

This proposes fixes and simplifications to those examples. Confirmed that they run successfully end-to-end with these changes (and produce some fun plots).

PR #519, opened 2025-02-10
Update AzureML docs

### LINKED ISSUE #515: Bug in AzureML cloud docs
In this page https://docs.rapids.ai/deployment/nightly/cloud/azure/azureml/, deployment works, but when running the first import 

```python
from azure.ai.ml import MLClient
```
we get this 

```
File /anaconda/envs/rapids/lib/python3.12/site-packages/azure/ai/ml/_schema/_data_import/data_import.py:9
      1 # ---------------------------------------------------------
      2 # Copyright (c) Microsoft Corporation. All rights reserved.
      3 # ---------------------------------------------------------
      4 
      5 # pylint: disable=unused-argument
      7 from marshmallow import post_load
----> 9 from azure.ai.ml._schema.core.fields import NestedField
     10 from azure.ai.ml._schema.job.input_output_entry import DatabaseSchema, FileSystemSchema
     11 from azure.ai.ml._utils._experimental import experimental

File /anaconda/envs/rapids/lib/python3.12/site-packages/azure/ai/ml/_schema/core/fields.py:19
     17 from marshmallow import RAISE, fields
     18 from marshmallow.exceptions import ValidationError
---> 19 from marshmallow.fields import _T, Field, Nested
     20 from marshmallow.utils import FieldInstanceResolutionError, from_iso_datetime, resolve_field_instance
     22 from ..._utils._arm_id_utils import AMLVersionedArmId, is_ARM_id_for_resource, parse_name_label, parse_name_version

ImportError: cannot import name '_T' from 'marshmallow.fields' (/anaconda/envs/rapids/lib/python3.12/site-packages/marshmallow/fields.py)
```

This seems to be a known issue https://github.com/Azure/azure-sdk-for-python/issues/39047 and it's technically fixed in https://github.com/Azure/azure-sdk-for-python/pull/39053, but I it doesn't look like there has been another release in conda, last one was in December https://anaconda.org/Microsoft/azure-ai-ml/files 

The PR in the conda forge feedstock is broken https://github.com/conda-forge/azure-ai-ml-feedstock/pull/29 not sure who can I ping to take a look at it. 

But they are cutting releases in pypi see https://pypi.org/project/azure-ai-ml/ 

I wonder if we can modify the .sh script to install those from pip? any suggestions?

ncclementi:
- Closes #515 

@jacobtomlinson this needs some input/answers hence opened as draft, but if you know the answer of some of this, feel free to either commit directly to the PR or leave a comment.


source/cloud/azure/azureml.md
```diff
@@ -28,7 +28,7 @@ The compute instance provides an integrated Jupyter notebook service, JupyterLab
 
 Sign in to [Azure Machine Learning Studio](https://ml.azure.com/) and navigate to your workspace on the left-side menu.
 
-Select **New** > **Compute instance** (Create compute instance) > choose a [RAPIDS compatible GPU](https://docs.rapids.ai/install/#system-req) VM size (e.g., `Standard_NC12s_v3`)
+Select **New** > **Compute instance** (Create compute instance) > choose an [Azure RAPIDS compatible GPU](https://docs.rapids.ai/deployment/stable/cloud/azure/) VM size (e.g., `Standard_NC12s_v3`)
```
ncclementi, line 31: Pointing this to the resources page of azure because it's more informative the the #system req page

source/cloud/azure/azureml.md
```diff
@@ -40,20 +40,19 @@ Choose "Provision with a creation script" to install RAPIDS and dependencies.
 Put the following in a local file called `rapids-azure-startup.sh`:
 
 ```bash
-#!/bin/bash
-
 sudo -u azureuser -i <<'EOF'
 source /anaconda/etc/profile.d/conda.sh
 conda create -y -n rapids \
-    {{ rapids_conda_channels }} \
+    -c rapidsai-nightly -c conda-forge -c nvidia \
```
ncclementi, line 46:
```suggestion
    {{ rapids_conda_channels }}
```

source/cloud/azure/azureml.md
```diff
@@ -40,20 +40,19 @@ Choose "Provision with a creation script" to install RAPIDS and dependencies.
...
     -c microsoft \
-    {{ rapids_conda_packages }} \
-    'azure-ai-ml>=2024.12' \
-    'azure-identity>=24.12' \
+    rapids=25.02 python=3.12 cuda-version=12.5 \
```
ncclementi, line 48:
```suggestion
   {{ rapids_conda_packages }} \
```

source/cloud/azure/azureml.md
```diff
@@ -40,20 +40,19 @@ Choose "Provision with a creation script" to install RAPIDS and dependencies.
...
 source /anaconda/etc/profile.d/conda.sh
 conda create -y -n rapids \
-    {{ rapids_conda_channels }} \
+    {{ rapids_conda_channels }}
     -c microsoft \
-    {{ rapids_conda_packages }} \
-    'azure-ai-ml>=2024.12' \
-    'azure-identity>=24.12' \
+   {{ rapids_conda_packages }} \
+    'azure-identity>=1.19' \
```
ncclementi, line 49: this one is more up to date in conda forge, so I replaced it.

source/cloud/azure/azureml.md
```diff
@@ -40,20 +40,19 @@ Choose "Provision with a creation script" to install RAPIDS and dependencies.
...
     ipykernel
 
 conda activate rapids
 
+pip install 'azure-ai-ml>=1.24'
```
ncclementi, line 54: this fix worked to fix the import issue related to marshmallow

source/cloud/azure/azureml.md
```diff
@@ -73,7 +72,7 @@ Once your Notebook Instance is `Running`, open "JupyterLab" and select the `rapi
 
 ## Azure ML Compute cluster
 
-Launch Azure's [ML Compute cluster](https://learn.microsoft.com/en-us/azure/machine-learning/how-to-create-attach-compute-cluster?tabs=python) to distribute your RAPIDS training jobs across a cluster of single or multi-GPU compute nodes.
+In the next section we will launch Azure's [ML Compute cluster](https://learn.microsoft.com/en-us/azure/machine-learning/how-to-create-attach-compute-cluster?tabs=python) to distribute your RAPIDS training jobs across a cluster of single or multi-GPU compute nodes.
```
ncclementi, line 75: I modify this to discuss this section. It's a bit unclear what to do here, because there is a link to some docs, so it seems like you have to do something, but you just need to follow the steps below.

source/cloud/azure/azureml.md
```diff
@@ -102,7 +101,9 @@ ml_client = MLClient.from_config(
 
 You will need to create a [compute target](https://learn.microsoft.com/en-us/azure/machine-learning/concept-compute-target?view=azureml-api-2#azure-machine-learning-compute-managed) using Azure ML managed compute ([AmlCompute](https://azuresdkdocs.blob.core.windows.net/$web/python/azure-ai-ml/0.1.0b4/azure.ai.ml.entities.html)) for remote training.
```
ncclementi, line 102: Should we say let's create or something that makes it clear to follow the next steps, instead of "You need to..." which reads as I need to do this before following along.

source/cloud/azure/azureml.md
```diff
@@ -102,7 +101,9 @@ ml_client = MLClient.from_config(
...
 
-Note: Be sure to check limits within your available region.
+```{note}
```
ncclementi, line 104: This was marked as note, but it's unclear where do we check this? I wasn't able to find where to check this

source/cloud/azure/azureml.md
```diff
@@ -146,7 +147,7 @@ You can define an environment from a [pre-built](https://learn.microsoft.com/en-
 
 Create your custom RAPIDS docker image using the example below, making sure to install additional packages needed for your workflows.
 
-```dockerfile
+```Dockerfile
```
ncclementi, line 150: I modify this to leave a particular comment. This section ("Custom Rapids Environment") is very confusing on what should we do, in particular because the following section ("Submit Rapids Training Jobs") in the `%bash` script we are creating the Dockerfile. 

I'm not super familiarized with how this work, I think it'll be good if someone that understand the flow can help clean this up or clarify what's needed.

source/cloud/azure/azureml.md
```diff
@@ -173,7 +174,7 @@ ml_client.environments.create_or_update(env_docker_image)
 
 Now that we have our environment and custom logic, we can configure and run the `command` [class](https://learn.microsoft.com/en-us/python/api/azure-ai-ml/azure.ai.ml?view=azure-python#azure-ai-ml-command) to submit training jobs.
 
-In a notebook cell, copy the example code from this documentation into a new folder.
+In a notebook cell, copy the example code from this documentation into a new folder, by running:
```
ncclementi, line 177: Do we need the creation of the Dockerfile in the previous step, given this here?  If not, then Should the code below "Now create the Environment, making sure to label and provide a description:" go below the `%bash` script run?

source/cloud/azure/azureml.md
```diff
@@ -232,7 +233,7 @@ returned_job = ml_client.jobs.create_or_update(command_job)
 
 After creating the job, go to [the "Experiments" page](https://ml.azure.com/experiments) to view logs, metrics, and outputs.
 
-Next, try performing a sweep over a set of hyperparameters.
+Next, perform a sweep over a set of hyperparameters. This will take (insert time?)
```
ncclementi, line 236: It's unclear how long things take, because in the notebook everything runs, but how are we sure things are running? What are we expecting here as output? 

Does this take a long time to run? Maybe we should disclaim this, because if this is your first time running it, it's unclear.

source/cloud/azure/azureml.md
jacobtomlinson: Sure sounds good

source/cloud/azure/azureml.md
jacobtomlinson: I think this is the quotas we were looking at the other day. Maybe rewording with "quota" would make this more understandable. Or if this is just duplication of other advice we could remove it.

source/cloud/azure/azureml.md
jacobtomlinson: There is definitely some duplication in here that will cause confusion. I don't think we need to create the `Dockerfile` in the bash section.

The steps for this kind of workflow are:
- Start from an interactive environment (could be your laptop or a compute instance)
- Create a compute cluster
- Create a software environment. This can be from a `Dockerfile`
- Submit a batch job that uses the environment to the compute cluster
- Wait for the results

source/cloud/azure/azureml.md
jacobtomlinson: Looking at the order of the page the previous cell creates the `Environment` object, which uses the `Dockerfile`. So I don't think we need to create it again in this bash script.

ncclementi: I see the bash script and some modifications to the Dockerfile in the section above where added by @jameslamb. James do you have any extra context about what seems like a duplication and what you've noticed while testing this? 

I personally like the idea of having the Dockerfile created by the bash script because it's easier, and in that case I'd move this part below the bash, right before the training. 
```python
env_docker_image = Environment(...)
```

source/cloud/azure/azureml.md
ncclementi: I went to the experiments page, but it was in "preparing" status for a long time, so I wasn't sure what I was looking for in there, and it seemed too long so I ended up cutting the experiment. 

Similar to the sweep step, I was not seeing anything so I was not sure what to expect. I think the whole job takes a long time, maybe @jameslamb knows, since you've run this before.

jameslamb: Not sure how long "long time" is. When I ran this in #484, I started seeing results on that page after about 20 minutes. And then they updated live as more experiments were run.

If I recall correctly, the entire job with the configuration in these docs took on the order of low-single-digit hours to complete.

source/cloud/azure/azureml.md
```diff
@@ -228,11 +224,19 @@ command_job = command(
...
+to [the "Experiments" page](https://ml.azure.com/experiments) to view logs, metrics, and outputs.
 
-Next, try performing a sweep over a set of hyperparameters.
+```{note}
+For reference this job took ~7min while using `size="Standard_NC6s_v3"` in the `gpu_compute` creation
+```
+
+![Screenshot of job under the test_rapids_mlflow experiment](../../images/azureml_returned_job_completed.png)
+
+Next, we can perform a sweep over a set of hyperparameters.
```
ncclementi, line 239: So I think we need to modify the parameters here, because this sweep creates 1000 jobs, I was able to complete ~5 in a span of 20 min, using `size="Standard_NC6s_v3"`. I'm not sure what's a more reasonable sweep but we should change this, any thoughts?

ncclementi:
This is ready for a final review. I left a comment about the sweep space, that we should definitely modify.

source/cloud/azure/azureml.md
jacobtomlinson: I think something small is totally fine. The goal here is to show how to dispatch the jobs, not how to do a good sweep. If you think it's best to do a small number of sweeps as an illustration then just make a note to say "this is a small example, in reality you would do more".

source/cloud/azure/azureml.md
```diff
@@ -254,8 +258,12 @@ sweep_job = command_job_for_sweep.sweep(
 
 # submit job
 returned_sweep_job = ml_client.create_or_update(sweep_job)
+returned_sweep_job
 ```
 
+Once the job is created, click on the details page provided in the output of `returned_sweep_job`, or go
```
jacobtomlinson, line 264: Oh great, the repr of the object gives you a link to the page?

jacobtomlinson (approved):

source/cloud/azure/azureml.md
ncclementi: Yes, to the specific job page under the experiment.

source/cloud/azure/azureml.md
ncclementi: Yes, I think I'll modify this and test it very quickly so it's also easier for us to test if things finish as expected.

source/cloud/azure/azureml.md
```diff
@@ -252,10 +256,20 @@ sweep_job = command_job_for_sweep.sweep(
     goal="Maximize",
 )
 
+# setting a very small limit of trials for demo purposes
+sweep_job.set_limits(
```
ncclementi, line 260: @jacobtomlinson I added this to get only 3 trials, given we have 3 GPU notes in the cluster created. each job took between 20-40 min, but I think this depends on the characteristics of the job, sometimes it's 10-15 min. I also put a note below about the times.  

If this is good, that's the last change I wanted to make. This should be good to go.

jacobtomlinson (approved):
Great thanks!

PR #524, opened 2025-02-12
add some clarification for the ec2 deployment process

taureandyernv:
updated the EC2 docs to reflect UI changes and possible challenges the user may face in deploying RAPIDS on an EC2 cluster as they work through the process


jacobtomlinson (approved):
Thanks @taureandyernv this looks great! I pushed a couple of commits to clean up the linting errors.

PR #526, opened 2025-02-12
Add related examples to snowflake doc

ncclementi:
I'm trying to add related examples to the snowflake docs, I thought that by just adding this

````text
```{relatedexamples}

```
````

They would show, but I don't see it locally. Is there an extra step that needs to be done to see them in the local build?

### CODE REVIEW

File: source/platforms/snowflake.md [hunk 1]
```diff
@@ -343,3 +343,7 @@ USE ROLE ACCOUNTADMIN;
 DROP ROLE CONTAINER_USER_ROLE;
 DROP EXTERNAL ACCESS INTEGRATION ALLOW_ALL_EAI;
 ```
+
```
jacobtomlinson, line 346: Let's add a title.

```suggestion
## Related Examples

```

jacobtomlinson (commented):
It's not showing because the tag doesn't match the page URI.

The filename is `platforms/snowflake.md` so the tag needs to be `platforms/snowflake`.  But it looks like the tag is `platform/snowflake`.

https://github.com/rapidsai/deployment/blob/27deca6ba711421c327b89e22f56d1371148e963/source/examples/rapids-snowflake-cudf/notebook.ipynb#L13

It looks like in the changes we made to make things more consistent recently some things may have been changed to not match the page URI. We should fix those up too.

source/platforms/snowflake.md
ncclementi: The magic behind 

```text
```{relatedexamples}

```
```

already includes adding the title. I had to fix all the tags to be able to get this one. So I include them as part of the fix of this PR.

jacobtomlinson (approved):

PR #530, opened 2025-02-15
Improve eks deployment docs

ncclementi:
Update EKS deployment docs clarifying ssh key steps, and including some notes that help during the deployment process.

### DIFF
--- a/source/cloud/aws/eks.md
+++ b/source/cloud/aws/eks.md
@@ -22,7 +22,7 @@ $ aws configure
 
 Now we can launch a GPU enabled EKS cluster with `eksctl`.
 
-```{notes}
+```{note}
 1. You will need to create or import public ssh-key to be able to execute the following command.
 In your aws console under `EC2` in the side panel under Network & Security > Key Pairs, you can create a
 key pair or import (see "Actions" dropdown) one you've created locally.
--- a/source/cloud/aws/eks.md
+++ b/source/cloud/aws/eks.md
@@ -37,7 +37,7 @@ $ eksctl create cluster rapids \
                       --node-type=g4dn.xlarge \
                       --timeout=40m \
                       --ssh-access \
-                      --ssh-public-key <public key ID> \  # Named assigned during creation of your key in aws console
+                      --ssh-public-key <public key ID> \  # Name assigned during creation of your key in aws console
                       --region us-east-1 \
                       --zones=us-east-1c,us-east-1b,us-east-1d \
                       --auto-kubeconfig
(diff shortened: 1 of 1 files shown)


source/cloud/aws/eks.md
```diff
@@ -20,24 +20,32 @@ $ aws configure
 
 ## Create the Kubernetes cluster
 
-Now we can launch a GPU enabled EKS cluster. First launch an EKS cluster with `eksctl`.
+Now we can launch a GPU enabled EKS cluster with `eksctl`.
+
+```{note}
+1. You will need to create or import public ssh-key to be able to execute the following command.
```
jameslamb, line 26:
```suggestion
1. You will need to create or import a public SSH key to be able to execute the following command.
```

I think the spelling `ssh-key` makes it seem like this might be a command, when really it's referring to just a keypair in general. Would you consider this rewording?

source/cloud/aws/eks.md
```diff
@@ -20,24 +20,32 @@ $ aws configure
...
 $ eksctl create cluster rapids \
-                      --version 1.29 \
+                      --version 1.30 \
                       --nodes 3 \
-                      --node-type=p3.8xlarge \
+                      --node-type=g4dn.xlarge \
                       --timeout=40m \
                       --ssh-access \
-                      --ssh-public-key <public key ID> \  # Be sure to set your public key ID here
+                      --ssh-public-key <public key ID> \  # Named assigned during creation of your key in aws console
```
jameslamb, line 40:
```suggestion
                      --ssh-public-key <public key ID> \  # Name assigned during creation of your key in aws console
```

source/cloud/aws/eks.md
```diff
@@ -20,24 +20,32 @@ $ aws configure
...
                       --region us-east-1 \
                       --zones=us-east-1c,us-east-1b,us-east-1d \
                       --auto-kubeconfig
 ```
 
 With this command, you’ve launched an EKS cluster called `rapids`. You’ve specified that it should use nodes of type `p3.8xlarge`. We also specified that we don't want to install the NVIDIA drivers as we will do that with the NVIDIA operator.
 
-To access the cluster we need to pull down the credentials.
+To access the cluster we need to pull down the credentials. Add `--profile <your-profile>` if you are not using the default profile.
```
jameslamb, line 48:
```suggestion
To access the cluster we need to pull down the credentials.
Add `--profile <your-profile>` if you are not using the default profile.
```

Could you move this sentence to its own line? That formatting for most markdown renderers is the same (will still be a paragraph with wrapping), but breaking it up across multiple lines makes editing a little easier and the git blame a little more useful as a source of "when did this last line change?".

jameslamb (approved):
Makes sense to me! I left just small, cosmetic suggestions, do what you want with them.

PR #537, opened 2025-04-03
Update Brev Instructions to include launchables

taureandyernv:
- [x] added Brev Launchables instructions
- [x] Updated existing Brev Instance instructions
- [ ] add new docker compose example


taureandyernv:
@jacobtomlinson , @ncclementi, @jameslamb , where would we want to have the example docker-compose.yaml?  I linked to the one we have, but happy to upload one and change it

source/cloud/nvidia/brev.md
```diff
@@ -4,41 +4,66 @@ review_priority: "p0"
...
+The [NVIDIA Brev](https://brev.dev/) platform provides you a one stop menu of available GPU instances across many cloud providers, including [Amazon Web Services](https://aws.amazon.com/) and [Google Cloud](https://cloud.google.com), with CUDA, Python, Jupyter Lab, all set up.
 
-You can get up and running with RAPIDS in a few quicks thanks to the Brev RAPIDS quickstart.
+## Brev Instance Setup 
+There are two ways that you can get up and running with RAPIDS in a few quicks thanks to the Brev RAPIDS quickstart:
+1. Brev GPU Instances - quickly get the GPU, across most clouds, to get your work done.
+2. Brev Launchables - quickly create one-click starting, reusable instances that you customized to your MLOps needs.
 
-## Setting up your instance
+### 1. Setting up your Brev GPU Instance
```
jacobtomlinson, line 14: I don't like adding numbers to these "GPU Instance" and "Launchables" headings. It feels like step one and step two, but they are alternative options. You wouldn't do both.

source/cloud/nvidia/brev.md
```diff
@@ -4,41 +4,66 @@ review_priority: "p0"
...
 
-Navigate to the [Brev console](https://console.brev.dev/) and click on "Create your first instance".
+1. Navigate to the [Brev console](https://console.brev.dev/) and click on "Create your first instance".
```
jacobtomlinson, line 16: Should this link to brev.nvidia.com now?

source/cloud/nvidia/brev.md
```diff
@@ -4,41 +4,66 @@ review_priority: "p0"
...
 
 ![Screenshot of the "Create your Instance" UI with an NVIDIA T4 GPU highlighted](/_static/images/platforms/brev/brev4.png)
 
-And hit "Deploy".
+5. And hit "Deploy".
 
 ![Screenshot of the instance creation summary screen with the deploy button highlighted](/_static/images/platforms/brev/brev5.png)
 
+### 2. Setting up your Brev Launchable
+1. Click Create Launchable
```
jacobtomlinson, line 37: Can we link to where people should be clicking this?

source/cloud/nvidia/brev.md
```diff
@@ -4,41 +4,66 @@ review_priority: "p0"
...
+1. Click **Compute** and select your GPU based on GPU type, number of GPUs, the cloud provider, and/or budget you have.  It is good to understand your GPU requirements before picking an instance, or use it to figure out which instance.  Once selected, click **Save**
```
jacobtomlinson, line 38: In the GPU instance section we are using quotes for the things people shoudl click.

> click on "Create your first instance".

Here we are using bold text

> Click **Compute**

Can we make things consistent one way or the other?

source/cloud/nvidia/brev.md
```diff
@@ -74,34 +99,41 @@ ssh <username>@<ip> -p 2222
...
+
+
+### Resources and tips
 
 - [Brev Docs](https://brev.dev/)
+- Please note: Git is not preinstalled in the RAPIDS container, but can be installed into the container when it is running using
+```
+apt update
+apt install git -y
+```
```
jacobtomlinson, line 139: This feels like pretty broad advice. Do we want to mention this in more places? Or should we consider adding `git` to the container if the requirement is strong enough that we need to document it?

source/cloud/nvidia/brev.md
```diff
@@ -4,41 +4,66 @@ review_priority: "p0"
...
+1. Click ** Container**.  When adding the **Container**, you can use the NVIDIA RAPIDS Container, use Docker Compose and edit our example yaml so that it preinstalls any additional conda or pip packages into your container before entry.  
+  1. If you just need standard NVIDIA RAPIDS install, Select **Container Mode > NVIDIA RAPIDS Container**. If using the Base Container, you may need to preinstall Jupyter.  If using the Notebooks Container, **Do not** Preinstall Jupyter, as that will break your instance.
+  1. If you need a customized environment, with additional packages on top of NVIDIA RAPIDSm use **Docker Compose** When using Docker Compose, you can upload a docker-compse yaml file.  Here is an example docker-compose file, **[docker/brev/docker-compose-nb-2412.yaml](https://github.com/clara-parabricks-workflows/single-cell-analysis-blueprint/raw/main/docker/brev/docker-compose-nb-2412.yaml)** that you can use as your base.  **Do not** Preinstall Jupyter when using that file, as it already will be installed.
```
jacobtomlinson, line 41: I think linking to the example is fine here. If we want to create a more bespoke example that walks you through doing a specific thing on brev then we should create another page in the workflow examples section.

jacobtomlinson (commented):
Thanks @taureandyernv. Some comments here that need fixing up.

source/cloud/nvidia/brev.md
```diff
@@ -4,41 +4,66 @@ review_priority: "p0"
 
 # NVIDIA Brev
 
-The [NVIDIA Brev](https://brev.dev/) platform provides you with GPU instances on cloud providers including [Amazon Web Services](https://aws.amazon.com/) and [Google Cloud](https://cloud.google.com) with CUDA, Python, Jupyter Lab, all set up.
+The [NVIDIA Brev](https://brev.dev/) platform provides you a one stop menu of available GPU instances across many cloud providers, including [Amazon Web Services](https://aws.amazon.com/) and [Google Cloud](https://cloud.google.com), with CUDA, Python, Jupyter Lab, all set up.
 
-You can get up and running with RAPIDS in a few quicks thanks to the Brev RAPIDS quickstart.
+## Brev Instance Setup 
+There are two ways that you can get up and running with RAPIDS in a few quicks thanks to the Brev RAPIDS quickstart:
```
ncclementi, line 10: The wording is a bit redundant, like using  quicks and quickstart
```suggestion
There are two ways that you can get up and running with RAPIDS in a few steps, thanks to the Brev RAPIDS quickstart:
```

source/cloud/nvidia/brev.md
```diff
@@ -4,41 +4,66 @@ review_priority: "p0"
...
+4. Configure your own instance.
 
 ![Screenshot of the "Create your Instance" UI with an NVIDIA T4 GPU highlighted](/_static/images/platforms/brev/brev4.png)
 
-And hit "Deploy".
+5. And hit "Deploy".
 
 ![Screenshot of the instance creation summary screen with the deploy button highlighted](/_static/images/platforms/brev/brev5.png)
 
+### 2. Setting up your Brev Launchable
```
ncclementi, line 36: All of the bullet points in the text bellow start with 1. can we make that consistent?

source/cloud/nvidia/brev.md
```diff
@@ -4,41 +4,66 @@ review_priority: "p0"
...
+  1. Click **Save**
+1. Click **Files** and add in any publicly avaialble single file or  github repository.  Click **Save**
+1. In **Ports**, please open ports **8888, 8787, and 8786**.  Name port 8888 `jupyter` so Brev can treat it as a Jupyter-Lab based instance and provide an **Open Notebook** button.  Click **Save**
+1. Name your Launchable, then Save your Launchable!
+1. Whenever you're ready to use your Launchable, Select your Launchable and hit **Deploy Launchable** 
+
 ## Accessing your instance
+There are a few ways to access your instance:
+1. Directly access Jupyter Lab from the Brev GUI
+1. Using the Brev CLI to connect to your instance....
```
ncclementi, line 51: Again all of the bullet points start with 1. and the CLI one has .... at the end.

source/cloud/nvidia/brev.md
```diff
@@ -4,41 +4,66 @@ review_priority: "p0"
...
+  1. Using Visual Studio Code
+  1. Using SSH via your terminal
+1. Access using the Brev tunnel
```
ncclementi, line 54: Some of this instructions look like should in Brev documentation, If any of these change in the future we will be out of sync, can we just link to their instructions?

source/cloud/nvidia/brev.md
```diff
@@ -74,34 +99,41 @@ ssh <username>@<ip> -p 2222
...
 You can verify that you have your requested GPU by running the `nvidia-smi` command.
 
 ![Screenshot of a notebook terminal running the command nvidia-smi and showing the NVIDIA T4 GPU in the output](/_static/images/platforms/brev/brev6.png)
 
-## Test RAPIDS
+### Test RAPIDS
 
 You can verify your RAPIDS installation is working by importing `cudf` and creating a GPU dataframe.
 
 ![Screenshot of a notebook cell importing and using cudf](/_static/images/platforms/brev/brev7.png)
```
ncclementi, line 128: Can we just have a code cell here that does the same, so the user can just copy and paste? A screenshot is not very helpful here.

ncclementi (commented):
I left couple of comments here, I think we are documenting more than what we need and some of the points made should be part of the Brev docs and we could just link to it.

jameslamb:
> `where would we want to have the example docker-compose.yaml`

If the Brev interface expects that you **paste in** a `docker-compose.yaml` file's contents (not that you provide a link to one that's hosted somewhere), then I think it'd make sense to inline that file directly into this documentation. So that folks reading the RAPIDS docs can just copy and past right from there. That'd also mean it could automatically get updated to new CUDA, Python, and RAPIDS versions via the templating in this repo.

The if the Brev interface expects that you **provide a link to a `docker-compose.yaml`**, then I think it'd be better to have a RAPIDS example file checked in in this repo, right next to the documentation describing it.

I'll defer to other reviewers for everything else, I can't review this more closely right now. Though based on a quick skim, I broadly agree with @ncclementi that much of this documentation doesn't look RAPIDS-specific. I'd expect the brev docs here to be pretty brief and to heavily rely on linking to existing brev documentation.

ncclementi:
> If the Brev interface expects that you paste in a docker-compose.yaml file's contents (not that you provide a link to one that's hosted somewhere), then I think it'd make sense to inline that file directly into this documentation

Brev requires a link.

taureandyernv:
> > `where would we want to have the example docker-compose.yaml`
> 
> If the Brev interface expects that you **paste in** a `docker-compose.yaml` file's contents (not that you provide a link to one that's hosted somewhere), then I think it'd make sense to inline that file directly into this documentation. So that folks reading the RAPIDS docs can just copy and past right from there. That'd also mean it could automatically get updated to new CUDA, Python, and RAPIDS versions via the templating in this repo.
> 
> The if the Brev interface expects that you **provide a link to a `docker-compose.yaml`**, then I think it'd be better to have a RAPIDS example file checked in in this repo, right next to the documentation describing it.
> 
> I'll defer to other reviewers for everything else, I can't review this more closely right now. Though based on a quick skim, I broadly agree with @ncclementi that much of this documentation doesn't look RAPIDS-specific. I'd expect the brev docs here to be pretty brief and to heavily rely on linking to existing brev documentation.

@jameslamb , I actually wanted the link to a more complete example docker file.  Where may i put that in this repo?

source/cloud/nvidia/brev.md
taureandyernv: Ah, these existed in the original file that i edited.  Would you rather i remove them?  @jacobtomlinson for thoughts, as original author

source/cloud/nvidia/brev.md
taureandyernv: i had to force this numbering to render correctly

source/cloud/nvidia/brev.md
taureandyernv: Gah, that was supposed to be `clicks`, but i can commit `steps`

source/cloud/nvidia/brev.md
taureandyernv: ah.  May I do Option 1 and Option 2?  I thought just bullets would be confusing.

jameslamb:
> Brev requires a link.
> ...
> @jameslamb , I actually wanted the link to a more complete example docker file. Where may i put that in this repo?

Ok, then I think right next to this new file you're adding is a good spot.

source/cloud/nvidia/brev.md
taureandyernv: sure

source/cloud/nvidia/brev.md
taureandyernv: true

source/cloud/nvidia/brev.md
taureandyernv: updated to all be consistant

source/cloud/nvidia/brev.md
taureandyernv: great point.  done!

jacobtomlinson (approved):
Things look pretty good here now. I've opened #539 to track the unchecked item on the list. I'm going to merge this as is and then @taureandyernv could you make a follow up with the docker compose pieces?

PR #541, opened 2025-04-08
Explicitly mention azureuser as the username for SSHing to an Azure VM

jacobtomlinson:
It took me a minute to figure out what the default username was for this VM type. Setting it explicitly here.


jameslamb (approved):
Yeah agree! I don't think this needs to be a a placeholder in this example... the default value is likely to be the real one people encounter.

PR #543, opened 2025-04-08
Ignore .ipynb_checkpoints in release checklist script

jacobtomlinson:
The last couple of releases I had to manually remove some `.ipynb_checkpoints` entries from the generated list. This PR updates the script so that they don't get included in the first place.


jameslamb (approved):
🎉

PR #553, opened 2025-05-07
Update Databricks wording to show we support the `15.x` and `16.x` ML GPU Runtime Environments

### LINKED ISSUE #552: Review supported Databricks ML environments
Currently we recommend using the `15.4 LTS ML (includes Apache Spark 3.5.0, GPU, Scala 2.12)` environment. As of writing this is the most recent LTS environment. However this environment includes `sklearn` version `1.3.0` which is older than `cuml` officially supports (`>=1.5`). As `cuml` is looking to make this a hard dependency it would be good if we supported a Databricks environment that has a newer `sklearn` version.

Historically we have only supported the most recent stable LTS environment, but we may need to break away from that in order to support `cuml` on Databricks going forwards.

Available environments that we should test:
- [x] 15.4 LTS ML (most recent stable LTS)
- [x] 16.0 ML
- [x] 16.1 ML
- [x] 16.2 ML
- [x] 16.3 ML
- [x] 16.4 LTS ML Beta

We can use [this simple notebook](https://gist.github.com/jacobtomlinson/e543eb40b7694d20e009798f011df6c4) to check imports and simple operations work as expected.

jacobtomlinson:
Closes #552

### DIFF
--- a/source/platforms/databricks.md
+++ b/source/platforms/databricks.md
@@ -42,7 +42,7 @@ To get started, navigate to the **All Purpose Compute** tab of the **Compute** s
 
 ![Screenshot of the Databricks compute page](../images/databricks-create-compute.png)
 
-In order to launch a GPU node uncheck **Use Photon Acceleration** and select `15.4 LTS ML (includes Apache Spark 3.5.0, GPU, Scala 2.12)` runtime version.
+In order to launch a GPU node uncheck **Use Photon Acceleration** and select any `15.x` or `16.x` ML runtime with GPU support. For example for long-term support releases you could select the `15.4 LTS ML (includes Apache Spark 3.5.0, GPU, Scala 2.12)` runtime version.
 
 The "GPU accelerated" nodes should now be available in the **Node type** dropdown.
 


source/platforms/databricks.md
```diff
@@ -42,7 +42,7 @@ To get started, navigate to the **All Purpose Compute** tab of the **Compute** s
 
 ![Screenshot of the Databricks compute page](../images/databricks-create-compute.png)
 
-In order to launch a GPU node uncheck **Use Photon Acceleration** and select `15.4 LTS ML (includes Apache Spark 3.5.0, GPU, Scala 2.12)` runtime version.
+In order to launch a GPU node uncheck **Use Photon Acceleration** and select any `15.x` or `16.x` ML runtime with GPU support. For example for long-term support releases you could select the `15.4 LTS ML (includes Apache Spark 3.5.0, GPU, Scala 2.12)` runtime version.
```
jameslamb, line 45:
```suggestion
In order to launch a GPU node uncheck **Use Photon Acceleration** and select any `15.x` or `16.x` ML runtime with GPU support.
For example for long-term support releases you could select the `15.4 LTS ML (includes Apache Spark 3.5.0, GPU, Scala 2.12)` runtime version.
```

I have a slight preference for breaking up markdown into 1-sentence-per-line, for easier view in `git` diffs and more-granular history in the git blame. The line break won't change the visual appearance of the rendered markdown at all.

Totally a personal-preference thing, do whatever you want with that suggestion.

jameslamb (approved):
One very very minor suggestion, but overall makes sense to me.

jacobtomlinson:
Thanks @jameslamb

PR #555, opened 2025-05-19
Morpheus, Kafka and RAPIDS deployment example

ncclementi:
This is WIP, so far putting manifests here


source/examples/rapids-morpheus-pipeline/notebook.md
```diff
@@ -0,0 +1,164 @@
...
+The core processing pipeline is deployed as a separate pod that, uses an image that can be built using the Dockerfile in
+the `pipeline-dockerfile` directory.
+
+- Runs the Morpheus nightly 25.06 conda build
+- Contains all pipeline and stage scripts `scripts/pipeline-dockerfile/*.py`
+- Processes the streaming data through the various stages
+
+(deploying-on-eks)=
+
+## Deploying on EKS
```
jacobtomlinson, line 82: We should probably `docref` the EKS page in here somewhere

source/examples/rapids-morpheus-pipeline/notebook.md
```diff
@@ -0,0 +1,164 @@
...
+
+The inference server is deployed using the NVIDIA Morpheus- Triton Inference Server docker image
+`nvcr.io/nvidia/morpheus/morpheus-tritonserver-models:25.02`.
+
+### Morpheus Pipeline Deployment (`k8s/morpheus-pipeline`)
+
+The core processing pipeline is deployed as a separate pod that, uses an image that can be built using the Dockerfile in
+the `pipeline-dockerfile` directory.
+
+- Runs the Morpheus nightly 25.06 conda build
```
jacobtomlinson, line 76: I'm apprehensive about putting specific versions in here that aren't part of our templating. This will go stale quickly. If it was a blog post I wouldn't be concerned because there is an expectation that it was written at some point in the past. However with docs guides there is an expectation for them to be kept up to date, so anything we can do to make things evergreen here we should do. Even if we refer readers to somewhere that they can find out what the latest release is.

source/examples/rapids-morpheus-pipeline/notebook.md
```diff
@@ -0,0 +1,164 @@
...
+In your aws console under `EC2` in the side panel under Network & Security > Key Pairs, you can create a
+key pair or import (see "Actions" dropdown) one you've created locally.
+
+2. If you are not using your default AWS profile, add `--profile <your-profile>` to the following command.
+```
+
+```console
+$ eksctl create cluster morpheus-rapids \
+    --version 1.32 \
+    --nodes 3 \
```
jacobtomlinson, line 103: We only end up using 2 GPUs in this demo, one for Triton and one for Morpheus. Maybe we should default this to `2`.

source/examples/rapids-morpheus-pipeline/notebook.md
```diff
@@ -0,0 +1,164 @@
...
+To access the cluster we need to pull down the credentials. Add `--profile <your-profile>` if you are not using the
+default profile.
+
+```console
+$ aws eks --region us-east-1 update-kubeconfig --name morpheus-rapids
+```
+
+### Deploy the Strimzi Operator
+
+Strimzi is an open-source project that provides a way to run Apache Kafka on Kubernetes. It
```
jacobtomlinson, line 122: Link to Strimzi?

source/examples/rapids-morpheus-pipeline/notebook.md
```diff
@@ -0,0 +1,164 @@
...
+simplifies the deployment and management of Kafka clusters by providing Kubernetes operators that
+handle the complex tasks of setting up and maintaining Kafka.
+
+We use `kubectl` to deploy teh operator. In our case we are deploying everything on the default
```
jacobtomlinson, line 126:
```suggestion
We use `kubectl` to deploy the operator. In our case we are deploying everything on the default
```

source/examples/rapids-morpheus-pipeline/notebook.md
```diff
@@ -0,0 +1,164 @@
...
+namespace, and the entire pipeline is design for that.
+
+```console
+$ kubectl create -f 'https://strimzi.io/install/latest?namespace=default'
```
jacobtomlinson, line 130: Specifying the namespace this way is a little unusual. Would this more idiomatic kubernetes method not work?

```suggestion
$ kubectl create -n default -f 'https://strimzi.io/install/latest'
```

source/examples/rapids-morpheus-pipeline/notebook.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+### Deploy the pipeline
+
+Get all the files in
+[https://github.com/rapidsai/deployment/source/examples/rapids-morpheus-pipeline/k8s](https://github.com/rapidsai/deployment/source/examples/rapids-morpheus-pipeline/k8s)
```
jacobtomlinson, line 136: Readers should be able to download these from the sidebar too if they choose. We might want to mention that.

source/examples/rapids-morpheus-pipeline/notebook.md
```diff
@@ -0,0 +1,164 @@
...
+
+Then do
+
+```console
+$ kubectl apply -f k8s --recursive
+```
+
+This will take ~15min to get all teh pods up and running, you will see for a while that the the `morpheus-pipeline` pod
```
jacobtomlinson, line 144:
```suggestion
This will take around 15min to get all the pods up and running, you will see for a while that the the `morpheus-pipeline` pod
```

source/examples/rapids-morpheus-pipeline/notebook.md
```diff
@@ -0,0 +1,164 @@
...
+fails and try to reconcile. This happens because the triton inference pod takes a while to get up and running.
+
+### Kafka UI: checking the pipeline results
+
+Once all the pods are running. You can check the input topic and the results topic in the kafka ui
```
jacobtomlinson, line 149:
```suggestion
Once all the pods are running. You can check the input topic and the results topic in the Kafka ui
```

jacobtomlinson (commented):
Thanks for this, a few comments:
- I fixed most of the build issues and pushed some commits. The other complaints are that TOC trees aren't up to date and there's a missing GIF which you already know about.
- Our extensions assume these examples are notebooks. Is there any reason why we can't make this a notebook where the `eksctl` and `kubectl` commands are code cells that can be executed if folks want to? This is how we've done other notebooks in the past.
- The tags displayed on the main examples page are loaded from the metadata of the first notebook cell. If we prefer to stick with markdown we would need to specify these some other way (frontmatter?) and update our extensions to support that.
- I noticed that the example doesn't talk about building the docker containers, we should definitely cover that.

source/examples/rapids-morpheus-pipeline/notebook.md
ncclementi: I was following their docs. I noticed when I tried running 
```
$ kubectl create  -f 'https://strimzi.io/install/latest'
```
which would be the default, it didn't deployed properly, and then the cluster later one was never up.  But I will double check. I might have have a typo

ncclementi:
Thanks @jacobtomlinson I'll go over all the review comments and move things to a notebook, it makes sense to be consistent and make the `eksctl` and `kubectl` code cells. 

> I noticed that the example doesn't talk about building the docker containers, we should definitely cover that.

I mentioned how to build this in the "Deployment components"  sections.  Both subsections "Kafka Producer Deployment" and "Morpheus Pipeline Deployment" say that it uses an image that can be built with the Dockerfiles. 

But I see I'm not providing clear links or instructions on how to do it. So I will make that better.

source/examples/rapids-morpheus-pipeline/notebook.md
jacobtomlinson: If it's necessary then that's totally fine, just wanted to check we were doing is an unusual way for a good reason.

source/examples/rapids-morpheus-pipeline/notebook.md
ncclementi: I though about it. I'm not sure how though because the EKS page is a simple rapids example.

ncclementi:
@jacobtomlinson I updated things, this might need a tiny bit more polishing.discussion. 

I've spent a decent amount of time, trying to figure out properly the retention policy on the messages, and while I was able to achieve a smaller build up, now we get some messages that are warnings on the pipeline that are like:

```text
%4|1748650600.175|OFFSET|rdkafka#consumer-1| [thrd:main]: network-traffic-input [1]: offset reset (at offset 2924865 
(leader epoch 0), broker 0) to offset END (leader epoch -1): fetch failed due to requested offset not available on the broker:
 Broker: Offset out of range
``` 

I think this is because the pipeline consumes messages at a way lower pace than what they are generated, which is probably because of the GPU we are using or network, unclear as this is very fast locally. I think it's not that relevant, but we should decide if we can live with that. Happy to chat about it.

jacobtomlinson:
Ok great, let's chat about it this week.

PR #560, opened 2025-05-21
Speed up unused images pre-commit hook

jacobtomlinson:
On my machine the `unused-images` pre-commit hook was taking ~35s. It seemed to be because it was traversing folders at the root level of the repo like `node_modules`. There were excludes for these but it wasn't working in the way I expected.

This PR updates the script to explicitly only check for images stored within `source/` and ensures they are referenced either in `source/` or the main `README.md`. This change reduces execution time to around 1 second.

### DIFF
--- a/scripts/unused_images.sh
+++ b/scripts/unused_images.sh
@@ -2,12 +2,12 @@
 
 set -e -u -o pipefail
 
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
🎉

PR #567, opened 2025-06-02
Enable md linting breakline and apply

### LINKED ISSUE #566: enable line-length break in `.mardownlint.json`
I noticed we have  `MD013: false` (see docs for MD013[ here](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md013---line-length)) set on our .markdownlint.json . Is tehre a historic reason for this setting? 

I personally find complicated to review PR that have a long paragraph with no break lines, specially when I want to suggest a change to specific text line. 

Can we enable this?

ncclementi:
- Closes #566

Notes:
- Chose 120 charcters becasue for md 80 seemed too short. 
- args: ["--fix"] not always work, as there are some limitations for long paragraphs, but the precommit tells you which line is that you need to shorten. 
- We are ignoring tables
- For some reason some console/bash blocks where been triggered by the linter (only 3 or 4 cases), I was able to ignore them by wrapping them 
```
<!-- markdownlint-disable -->
console block here
<!-- markdownlint-enable -->
``` 
- I applied the linting to each doc that needed it separately, just in case we need to revert something. When squashing delete the message.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,10 +102,14 @@ user experience and there may already be a suitable root namespace for the tag y
 By default tags are styled with RAPIDS purple backgrounds and white text. They also have a `0.5em` left hand border to
 use as an accent that is also purple by default which can be styled for a two-tone effect.
 
+<!-- markdownlint-disable -->
+
 <div style="width: 100%; text-align: center;">
-<img alt="Diagram showing the tag and css side-by-side with arrows to show color sets the text, background-color sets
-the background and border-left sets the accent" src="source/images/theme-tag-style.png" style="max-width: 450px;" />
+<img alt="Diagram showing the tag and css side-by-side with arrows to show color sets the text, background-color sets the background and border-left sets the accent"
+src="source/images/theme-tag-style.png" style="max-width: 450px;" />
 </div>
+
+<!-- markdownlint-enable -->
 
 This can be overridden for each tag by adding a new class with the format `.tag-{name}` to `source/_static/css/custom.css`.
 For example the Scikit-Learn logo is orange and blue with grey text, so the custom CSS sets an orange background with a
--- a/README.md
+++ b/README.md
@@ -36,8 +36,7 @@ uv lock --upgrade
 
 ## Writing
 
-Content in these docs are written in markdown using the [MyST Sphinx
-extension](https://myst-parser.readthedocs.io/en/v0.15.1/syntax/syntax.html).
+Content in these docs are written in markdown using the [MyST Sphinx extension](https://myst-parser.readthedocs.io/en/v0.15.1/syntax/syntax.html).
 
 ### Custom admonitions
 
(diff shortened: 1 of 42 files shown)


source/cloud/azure/azureml.md
```diff
@@ -4,31 +4,47 @@ review_priority: "p0"
...
+RAPIDS can be deployed at scale using [Azure Machine Learning
+Service](https://learn.microsoft.com/en-us/azure/machine-learning/overview-what-is-azure-machine-learning) and can be
+scaled up to any size needed.
 
 ## Pre-requisites
 
-Use existing or create new Azure Machine Learning workspace through the [Azure portal](https://learn.microsoft.com/en-us/azure/machine-learning/how-to-manage-workspace?tabs=azure-portal#create-a-workspace), [Azure ML Python SDK](https://learn.microsoft.com/en-us/azure/machine-learning/how-to-manage-workspace?tabs=python#create-a-workspace), [Azure CLI](https://learn.microsoft.com/en-us/azure/machine-learning/how-to-manage-workspace-cli?tabs=createnewresources) or [Azure Resource Manager templates](https://learn.microsoft.com/en-us/azure/machine-learning/how-to-create-workspace-template?tabs=azcli).
+Use existing or create new Azure Machine Learning workspace through the [Azure
+portal](https://learn.microsoft.com/en-us/azure/machine-learning/how-to-manage-workspace?tabs=azure-portal#create-a-workspace),
```
jacobtomlinson, line 14: This has wrapped within a link, I'm not sure this is expected.

README.md
```diff
@@ -79,21 +85,32 @@ The notebook gallery extension uses cell tags to organize and cross-reference fi
...
 ##### Styling
 
-By default tags are styled with RAPIDS purple backgrounds and white text. They also have a `0.5em` left hand border to use as an accent that is also purple by default which can be styled for a two-tone effect.
+By default tags are styled with RAPIDS purple backgrounds and white text. They also have a `0.5em` left hand border to
+use as an accent that is also purple by default which can be styled for a two-tone effect.
 
 <div style="width: 100%; text-align: center;">
-<img alt="Diagram showing the tag and css side-by-side with arrows to show color sets the text, background-color sets the background and border-left sets the accent" src="source/images/theme-tag-style.png" style="max-width: 450px;" />
+<img alt="Diagram showing the tag and css side-by-side with arrows to show color sets the text, background-color sets
+the background and border-left sets the accent" src="source/images/theme-tag-style.png" style="max-width: 450px;" />
```
jacobtomlinson, line 108: This has wrapped within an HTML tag which will break the HTML

jacobtomlinson (commented):
Thanks @ncclementi. I'm not convinced this is an improvement though and I'm currently -1 on merging it.

Enforcing an arbitrary line length in markdown breaks text wrapping in other tools like editors, GitHub previews, etc with trailing words. Depending on folk's preferred font size, and spacing settings it causes things to wrap in weird places instead of just letting the editor/renderer handle the wrapping.

I do agree with @jameslamb's comment about having a preference for one sentence per line https://github.com/rapidsai/deployment/pull/553#discussion_r2077744473, but enforcing a character length isn't quite the same thing.

I feel like `markdownlint` is a much less refined tool than something like `black` and it often makes a mess of things. There are many examples in this PR where it has broken links and HTML over multiple lines which changes their semantic meaning.

ncclementi:
After talking, we decided that markdown lint does not do what we need in a way that doesn't introduce other problems, so we are not going to pursue this path.

Issue #573, opened 2025-06-05
Using UpperCamelCase for Kubernetes resources


jacobtomlinson:
The Kubernetes [documentation style-guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-upper-camel-case-for-api-objects) encourages the use of UpperCamelCase (also known as PascalCase) when referring to Kubernetes resources.

It's a little nitpicky but I think it would be good to be consistent with this in our own documentation. For example we regularly writer `pods` when we should write `Pods`.

### Examples in our docs

https://github.com/rapidsai/deployment/blob/104ebd31cdc232e7b4e15e7b32488935b5ab99c5/source/platforms/kubeflow.md?plain=1#L132
https://github.com/rapidsai/deployment/blob/d1264c7ee5b08e29ed63b68872f784627b5022de/source/tools/kubernetes/dask-helm-chart.md?plain=1#L47
https://github.com/rapidsai/deployment/blob/d1264c7ee5b08e29ed63b68872f784627b5022de/source/guides/index.md?plain=1#L57

### Guidance Examples

Guidance form the [style-guide](https://kubernetes.io/docs/contribute/style/style-guide/#use-upper-camel-case-for-api-objects)

Do | Don't
-- | --
The HorizontalPodAutoscaler resource is responsible for ... | The Horizontal pod autoscaler is responsible for ...
A PodList object is a list of pods. | A Pod List object is a list of pods.
The Volume object contains a hostPath field. | The volume object contains a hostPath field.
Every ConfigMap object is part of a namespace. | Every configMap object is part of a namespace.
For managing confidential data, consider using the Secret API. | For managing confidential data, consider using the secret API.

ncclementi:
Do things like `configMap` in yaml files need to be changed for `ConfigMap`?

ncclementi:
I'll re-open this, because I'm not sure the fixes I made are indication of completion.

jacobtomlinson:
I think if we update the README with this style guide info and progressively fix things over time then we could probably close this.

PR #574, opened 2025-06-05
Address K8s style guide

ncclementi:
- Partially resolves https://github.com/rapidsai/deployment/issues/573

Mostly replaced pod and pods for Pod and Pods, and one appearance of config map for ConfigMap


ncclementi:
I'm not sure how to capture or find more, but we can check them little by little.

Issue #584, opened 2025-06-30
Upgrading to CUDA 12.9 (in RAPIDS 25.08)


jakirkham:
In RAPIDS 25.08, we are updating CUDA 12 from 12.8 to 12.9. There may be a few touch-ups needed to capture this in the deployment docs

Also worth noting that RAPIDS 25.06 is installable with and was tested on CUDA 12.9. So mentioning CUDA 12.9 even with RAPIDS 25.06 is ok. We only started building with CUDA 12.9 in RAPIDS 25.08

ncclementi:
Closing as completed

PR #585, opened 2025-07-01
Bump nightly tags to CUDA 12.9

jacobtomlinson:
Towards #584 (we will need to bump stable as part of the release cc @AyodeAwe)


ncclementi (approved):
LGTM 🚀

Issue #586, opened 2025-07-15
Remove uses of `cuspatial`


jameslamb:
## Description

RAPIDS 25.04 was the final release to include `cuspatial` and `cuproj`:

* https://github.com/rapidsai/build-planning/issues/197
* https://docs.rapids.ai/notices/rsn0045/

The code samples here using those libraries should be removed or updated to work without them.

## Benefits of this work

* ensures that examples work when users try them

## Acceptance Criteria

* 0 remaining uses of `cuspatial` or `cuproj` in code samples OR those samples explicitly pin to older RAPIDS releases

## Approach

Search like this:

```shell
git grep -i -E 'cuproj|cuspatial'
```

Update and test whatever's found. As of this writing, that's the following.

https://github.com/rapidsai/deployment/blob/86bd08e0328098f55e92a03cc03e09059a6974d6/source/examples/rapids-ec2-mnmg/notebook.ipynb?short_path=1bfd04c#L302

https://github.com/rapidsai/deployment/blob/86bd08e0328098f55e92a03cc03e09059a6974d6/source/examples/rapids-autoscaling-multi-tenant-kubernetes/notebook.ipynb?short_path=86be9cf#L998

https://github.com/rapidsai/deployment/blob/86bd08e0328098f55e92a03cc03e09059a6974d6/source/examples/xgboost-gpu-hpo-mnmg-parallel-k8s/notebook.ipynb?short_path=8fa7d90#L244

## Notes

N/A

ncclementi:
I wonder what is the right path here, in the three notebooks above cuspatial is used to compute the harvesine distance between pickup and dropoff (lat,lon) points 

```python
def haversine_dist(df):
    import cuspatial

    pickup = cuspatial.GeoSeries.from_points_xy(
        df[["pickup_longitude", "pickup_latitude"]].interleave_columns()
    )
    dropoff = cuspatial.GeoSeries.from_points_xy(
        df[["dropoff_longitude", "dropoff_latitude"]].interleave_columns()
    )
    df["h_distance"] = cuspatial.haversine_distance(pickup, dropoff)
    df["h_distance"] = df["h_distance"].astype("float32")
    return df
```

1. We could wither not get this feature and use a linear distance or a different example. 
2. Leave them as they are and since we are planning on adding dates to workflows (xref: https://github.com/rapidsai/deployment/issues/576), which means we are not guaranteeing the code runs.

PR #587, opened 2025-07-16
remove cuspatial references in docs

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/197

The last RAPIDS release with `cuspatial` packages was 25.04. This removes references to it in docs.

### How I tested this

Looked for references like this:

```shell
git grep -i -E 'cuproj|cuspatial'
```

### Other work

There are some uses of `cuspatial` in a few example notebooks here. Those should be updated or removed. Documented that in #586


ncclementi (approved):
Thank you @jameslamb

Issue #407, opened 2024-08-06
Remove L4 page


jacobtomlinson:
Currently we have a [page dedicated to using L4 GPUs on Google Cloud](https://docs.rapids.ai/deployment/nightly/guides/l4-gcp/).

We have to document this separately to creating other [Google Cloud VMs](https://docs.rapids.ai/deployment/nightly/cloud/gcp/compute-engine/) because the NVIDIA GPU-Optimized VMI doesn't allow you to select an L4 (or H100 for that matter).

It would be great to consolidate these pages into a single process if the VMI ever supports this.

ncclementi:
I'm not sure why this one wasn't closed with [Remove gcp-l4 page #588](https://github.com/rapidsai/deployment/pull/588), but closing now. If I'm missing something feel free to re-open

PR #588, opened 2025-07-21
Remove gcp-l4 page

ncclementi:
- [x] Closes https://github.com/rapidsai/deployment/issues/407


jacobtomlinson (commented):
Thanks @ncclementi.

Could we use something like [sphinx-reredirects](https://pypi.org/project/sphinx-reredirects/) to redirect this URL to the GCP page?

jacobtomlinson (approved):
Thanks!

ncclementi:
Good catch, I'm not sure if this works on PRs though. 

I had to difget a bit with how to specify the paths but I still can't get it to work. on the PR. 
The redirect of 
 https://rapids-deployment--588.org.readthedocs.build/en/588/guides/l4-gcp  converting to this
 https://rapids-deployment--588.org.readthedocs.build/guides/l4-gcp  and when it removes /en/588 then it leads to page note found
 
 Idk if it's just because of how read the docs is, or if we need to work with relative paths.

jacobtomlinson:
Let's merge this and see how things look in the nightly. I'll open another issue to track fixing the redirects. We can close it immediately if this does indeed work.

PR #597, opened 2025-08-12
Improve unused image check performance on Mac

jacobtomlinson:
The unused images pre-commit check runs slow on Mac, but not on linux.

I usually run things on my workstation which has `grep` 3


ncclementi (approved):
LGTM

Issue #534, opened 2025-02-20
Write guidelines of use of`$` in bash/console commands


ncclementi:
In our docs we have multiple places where we have `$` in front of a bash command and in others we don't.  When present it's not copied when either selecting and using the copy to clipboard shortcut.  But the `$` will cause problems in multi line commands. This was fixed across the docs I believe, but we should write this as a guideline for future reference. 

I would prefer we don't use it at all for simplicity, but this is a problem when we have a one liner with an output. For example: 

````text
```bash
$ echo $PATH
/some/path
```
````

In our docs, when you copy the cell above, you'll get only the one line that has `$`. 

The one thing it's not clear is when we use `console` and when we use `bash` or if it doesn't matter. Ideally we should be consistent too. 

TODO:
- [ ] Decide weather to use `$` only having one liners and displaying the output, or not use it and modify across the docs 
- [ ] Determine when we use `console` vs `bash` and if they are the same, choose one for consistency 
- [ ]  Write guidelines in the README under the "Writing" section.

**Comment by ncclementi** [comment]
> You should use them if you're using the console highlighting (for commands and output) and not if you're using the bash highlighting (commands without output). Although not many folks know about this and often get things mixed up.

I notice multiple places were we have one liners that look like this, 

````text
```console
$ gcloud init
```
````

are we saying in this case we should do? 

````text
```bash
gcloud init
```
````

> With console it highlights the line with the $ as bash and the following lines as text

This only seems to happen at the markdown level, for example I see it here: 
https://github.com/rapidsai/deployment/blob/00321049d5040ada4bc541da6727209cccb2ca20/source/guides/colocate-workers.md?plain=1#L39-L47

But in the rendering this it looks like this, is it supposed to have a different highlight in the rendering? 

![Image](https://github.com/user-attachments/assets/f579b764-2b38-48a5-922e-19e74e5751e1)

jacobtomlinson:
> I notice multiple places were we have one liners that look like this

In theory yes that should be `bash` and not use the `$`. If there are cases like that we should fix them. However, if there are a whole load of `console` examples and one of them is a one-liner it might be weird to have one `bash` mixed into a bunch of `console` blocks.

> is it supposed to have a different highlight in the rendering?

[...]

Issue #283, opened 2023-08-24
[DOC] fix code blocks in md files to correctly identify code for copying


taureandyernv:
In several documents pages, such as https://docs.rapids.ai/deployment/nightly/cloud/aws/eks/, the conversion from markdown to sphinx docs breaks the multiline copy function in the `console` code block.  When the user attempts to copy the code command from the block using the button or parsing the markdown, they only get the first line of the command instead of the complete multiline command.   This does not happen in the Github previews

### Example from https://docs.rapids.ai/deployment/nightly/cloud/aws/eks/: 
Github preview copy command results: 
```
$ eksctl create cluster rapids \
`                      --version 1.24 \
                      --nodes 3 \
                      --node-type=p3.8xlarge \
                      --timeout=40m \
                      --ssh-access \
                      --ssh-public-key <public key ID> \  # Be sure to set your public key ID here
                      --region us-east-1 \
                      --zones=us-east-1c,us-east-1b,us-east-1d \
                      --auto-kubeconfig \
                      --install-nvidia-plugin=false
```
Deployments EKS webpage copy command results
```
eksctl create cluster rapids \
```

PR #603, opened 2025-08-25
Standardize bash vs console usage and fix copy multiline

ncclementi:
Note: Check conflicts when merging since there are multiple PRs ongoing touching several files.

- Closes https://github.com/rapidsai/deployment/issues/534
- Closes https://github.com/rapidsai/deployment/issues/283

This PR
- Adds this to the config.py fixes the multiline copying issue in `.md`.

```bash
copybutton_line_continuation_character = "\\"
```

In `.md` files  (it's ok in `.ipynb` for some reasons) we can't add comments inline of a multiline bash command like this, but seems a reasonable thing to avoid. 

````text
```console
$ docker run --gpus all --rm -it \
    --shm-size=1g --ulimit memlock=-1 \
    -p 8888:8888 -p 8787:8787 -p 8786:8786 \ # this will break the copying, omitting anything after the comment
    rapidsai/notebooks:25.10a-cuda12.9-py3.12
```
````

- Standardizes the usage of console and bash. Console is used in anything that is a command line block, and uses bash for anything that is a bash script, as in `.sh` file. 
- Adds guidance to the readme on what to use when. 

Out of scope: 

- https://github.com/rapidsai/deployment/issues/602

### CODE REVIEW

File: source/cloud/azure/aks.md [hunk 1]
```diff
@@ -91,8 +91,8 @@ $ az extension add --name aks-preview
 
 `````
 
-```bash
-az aks nodepool add \
+```console
```
jacobtomlinson, line 94: No output should be bash

jacobtomlinson (changes requested):
Thanks for starting this.

I found myself leaving the same comment so instead of doing that on every instance I'll just write it here.

We only want to use `console` when we are showing output from the command. Otherwise we should just use `bash`. Many markdown linters will complain about console blocks without output.

```mermaid
flowchart TD
    A[Single line command or multi-line script]
    A -->|multi| B[```bash]
    A -->|single| C[Does the command have output?]
    C -->|no| B
    C -->|yes| D[```console]
```

source/cloud/aws/ec2-multi.md
```diff
@@ -20,7 +20,7 @@ Install the AWS CLI tools following the [official instructions](https://docs.aws
 
 Also install `dask-cloudprovider` and ensure you select the `aws` optional extras.
 
-```console
+```bash
 $ pip install "dask-cloudprovider[aws]"
```
jacobtomlinson, line 24: When it's `bash` you omit the `$`.

source/cloud/nvidia/brev.md
```diff
@@ -166,6 +166,9 @@ print(gdf)
 - Please note: Git is not preinstalled in the RAPIDS container, but can be installed into the container when it is running using
 
 ```bash
-apt update
-apt install git -y
+$ apt update
+```
+
+```bash
```
jacobtomlinson, line 172: This could be a multi-line block. Was there a reason for splitting it up?

jacobtomlinson (commented):
Thanks for going over this again, I know it's a bit of a tedious one.

Unfortunately I think you also need to go through and remove the `$` from the bash blocks too.

source/cloud/aws/ec2-multi.md
ncclementi: Is there a particular reason for that? Because we have examples in the docs of is using `$` with bash 

https://github.com/rapidsai/deployment/blob/03c86b0b846e5a0e7e36fa4e0bf624b7f0c0139f/source/tools/kubernetes/dask-helm-chart.md?plain=1#L117-L123

For reference, when we copy this with the button it only copies the command.

source/cloud/nvidia/brev.md
ncclementi: The problem with that is that the copy button copies everything and in my case (a personal preference) I rather have the choice of executing one command at the time. 

But I can revert it if you have a strong preference for executing them all at once.

Issue #592, opened 2025-07-30
Fix page redirects


jacobtomlinson:
@ncclementi discovered in #588 that the redirects extension is not working. It's also behaving strangely in PR preview builds.

We should explore this further, especially as we want to remove older pages we should redirect them somewhere useful.

ncclementi:
I think re-directs are still broken, and can't quite pinpoint what's happening. 

1. I'm seeing something very strange for the l4 page 

If I go to https://docs.rapids.ai/deployment/nightly/guides/l4-gcp/ it actually pulls the page, but I recall deleting it in the PR from 588, so I can't understand what's happening. 

2. https://docs.rapids.ai/deployment/nightly/platforms/brev-dev 

Leads to a page not found.

PR #607, opened 2025-08-28
Fix redirects to match the correct build

ncclementi:
Closes https://github.com/rapidsai/deployment/issues/592

I think the reason why this work and before it wasn't, is because what is on this PR matches the sphinx's builder `sphinx-autobuild source build/html` where outputs are *.html (e.g., platforms/brev-dev.html) rather than directories. 

That being said. I'm still unsure if we want to keep this redirects, as I don't see how someone will purposely navigate to `platforms/brev-dev`. They'll have to type it or have it booked marked. But that's a separate discussion, happy to keep start a new issue if we want to discuss this.


jacobtomlinson:
Oh I think I know what is going on here. In the main builds we are using the `dirhtml` builder. So `platforms/brev-dev.md` becomes `platforms/brev-dev/index.html`. But I think the dev/CI builds are using the `html` builder where `platforms/brev-dev.md` becomes `platforms/brev-dev.html`.

I think the redirects config are correct but the build preview commands are wrong.

I can confirm the brev redirect works for me if I run `sphinx-autobuild -b dirhtml source build/html`

jacobtomlinson:
Sure but that's a different problem where it doesn't know about the base URL. We can [configure the base url](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_baseurl) in `conf.py`, but we only need to do that for the final builds, not development ones.

ncclementi:
Oh I see, is there a way of configuring the base url to be either stable/ or nightly/ depending on the situation? I haven't investigated this, but if you have any pointers I can try to work this one out.

ncclementi:
To check it works locally, build using `dirhtml`

```
sphinx-autobuild -b dirhtml source build/html 
```

ncclementi:
Updated the README to do the autobuild with dirhtml.

PR #611, opened 2025-09-16
Update instructions to UCXX

pentschev:
UCX-Py is being archived, therefore instructions need to be updated to refer/use UCXX.


jacobtomlinson (approved):
Thanks Peter!

PR #620, opened 2025-10-06
Update Coiled conda channels to support nightly

jacobtomlinson:
Currently only stable works, this PR allows for easier testing of nightly. Found while testing #618


ncclementi (approved):
LGTM, minor comment to fix in case is possible.

PR #621, opened 2025-10-06
Fix inconsistent VM type in EKS page

jayavenkatesh19:
Fixes mismatch between `g4dn.xlarge` in the cluster creation command and `p3.8xlarge` described in the text underneath.


ncclementi (approved):
Thank you @jayavenkatesh19

PR #626, opened 2025-10-09
Update brev launchable instructions

ncclementi:
- Make clear how to setup things via docker compose with rapids notebook, add a few screenshots and a template of a docker compose yaml. 
 
-  Note that I'm only exposing the Jupyter port (8888) to simplify things, but if dask is desired, I think we need to sort how to expose the ports properly because the scheduler one it always shows unhealthy, while the dashboard one turns healthy when creating a Client (this makes sense).  

@jacobtomlinson I think we can pair on the scheduler port at some point and improve the docs for this case. 

While I'm out push as much as you need to the PR to get this one across the line.

### DIFF
--- a/source/cloud/nvidia/brev.md
+++ b/source/cloud/nvidia/brev.md
@@ -69,7 +69,9 @@ You can read more about Brev Launchables in the [Getting Started Guide](https://
         want to work with. Note, you will not be able to resize the instance once created.
       - Create a name for your launchable, and deploy.
    2. **Docker Compose**: For a custom container that you can tailor to your needs.
-      - You can provide a `docker-compose.yaml` via url o from a local file. The structure should be as follow
+      - You can provide a `docker-compose.yaml` via url o from a local file. In the following template, make sure to
+        replace `<name_of_your_github_repo>` in the `volumes` path, with the name of your repository if you have one. Otherwise,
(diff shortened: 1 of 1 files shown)


source/cloud/nvidia/brev.md
```diff
@@ -49,31 +49,65 @@ Brev Launchables are shareable environment configurations that combine code, con
...
-7. Select your desired compute environment. Make sure you select sufficient disk size to download the datasets you want
-   to work with. Note, you will not be able to resize the instance once created.
-8. Create a name for your launchable, and deploy 🚀!
+   1. **"Featured Container"** and select the "NVIDIA RAPIDS" container: For a ready to go environment with the
+      entire RAPIDS stack and Jupyter configured.
+      - Select your desired compute environment. Make sure you select sufficient disk size to download the datasets you
+        want to work with. Note, you will not be able to resize the instance once created.
+      - Create a name for your launchable, and deploy.
+   2. **Docker Compose**: For a custom container that you can tailor to your needs.
+      - You can provide a `docker-compose.yaml` via url o from a local file. The structure should be as follow
```
jacobtomlinson: I would explicitly instruct people here to "Replace `<name_of_your_github_repo>` with the name of your GitHub repo". Otherwise people might not read it fully.

PR #638, opened 2025-11-20
Add a Workflow Example for Geospatial Analysis using RAPIDS

jayavenkatesh19:
Adding a notebook here which consists of an end-to-end workflow example to load and preprocess Sentinel-2 data, train a Random Forest model to perform LULC classification and perform inference on an unseen Sentinel-2 tile.


jacobtomlinson (commented):
This is looking really nice! I left a few comments throughout (I'm trying out [conventional comments](https://conventionalcomments.org/), let me know what you think).

A few higher level thoughts:

The chunk of text above each code cell feels a little robotic. I'm guessing you used AI to generate some portion of these. It feels a little like that code comment trap where you just describe what the code does. I do appreciate some amount of that because the code is necessarily terse, lots of geo data analysis code is. But I think what it is missing is "why". You start out at the beginning with some high level goals. It would be nice to keep a narrative thread throughout explaining why you are doing each things and how each step is bringing us closer to that goal.

The notebook feels like it ends abruptly. There is no conclusion or recap of what was done. I think you need to wrap the post up in some way, probably by recapping what was done and then talking a little more about what could be improved or done next.

ncclementi (approved):
LGTM

Issue #54, opened 2022-10-31
Document deploying RAPIDS with SLURM


jacobtomlinson:
We have existing SLURM documentation here https://rapids.ai/hpc. We should migrate this to the deployment documentation here.

jacobtomlinson:
This gist is a good starting point https://gist.github.com/jacobtomlinson/d101991adefc1e85e3dafd67ff7f1fb1

ncclementi:
Closing since https://github.com/rapidsai/deployment/pull/663

PR #663, opened 2026-02-11
Update HPC documentation page

jayavenkatesh19:
Towards #54 

Replaces the legacy HPC page (outdated Dask with CUDA 11.0.3 content) with a modern guide covering single-GPU RAPIDS deployment on SLURM clusters. 

Consists of the following sections:

- SLURM basics: Partitions, interactive vs bash jobs, srun/sinfo/sbatch flags, session persistence with tmux
- Lmod RAPIDS module creation
- Apptainer/Enroot+Pyxis RAPIDS containers
- Single GPU cudf.pandas example. 

Multi-GPU workloads, cloud SLURM (Coreweave) will be added to this page in further iterations.


source/hpc.md
```diff
@@ -4,129 +4,311 @@ review_priority: "index"
 
 # HPC
 
-RAPIDS works extremely well in traditional HPC (High Performance Computing) environments where GPUs are often co-located with accelerated networking hardware such as InfiniBand. Deploying on HPC often means using queue management systems such as SLURM, LSF, PBS, etc.
+RAPIDS can be deployed on HPC clusters managed by [SLURM](https://slurm.schedmd.com/).
```
ncclementi: I think we can keep some of the info we had before, we don't have to remove this part. We can keep it 

"RAPIDS works extremely well in traditional HPC (High Performance Computing) environments where GPUs are often co-located with accelerated networking hardware such as InfiniBand."

source/hpc.md
```diff
@@ -4,129 +4,311 @@ review_priority: "index"
...
-dask-cuda-worker \
-    --rmm-pool-size 14GB \
-    --scheduler-file "$LOCAL_DIRECTORY/dask-scheduler.json"
+[Environment modules](https://modules.readthedocs.io/) are the standard way
+to manage software on HPC clusters. We'll create a
+[conda](https://docs.conda.io/) environment containing both CUDA and RAPIDS,
+then wrap it in an [Lmod](https://lmod.readthedocs.io/) module file so it can
+be loaded with a single command.
+
+We use conda here because it handles the CUDA toolkit and RAPIDS dependencies
```
ncclementi: This is true, but now for cudf and cuml pip handles this too. So I'm not sure if we want to write it necessarily that way. But if we are installing all rapids then I'm ok leaving this message like this.

source/hpc.md
```diff
@@ -4,129 +4,311 @@ review_priority: "index"
...
+    {{ rapids_conda_channels }} \
+    {{ rapids_conda_packages }}
+```
+
+#### Create the module file
 
-#SBATCH -J dask-client
-#SBATCH -n 1
-#SBATCH -t 00:10:00
+Place a module file in your cluster's module path so that users can load
```
ncclementi: Are we assuming conda is installed?  We should add how to install conda, as an instruction to make the content self contained. 

 Across our docs, we usually recommend miniforge.

source/hpc.md
```diff
@@ -4,129 +4,311 @@ review_priority: "index"
...
+the environment. Replace `<path to miniconda3>` with the absolute path to
```
ncclementi: Let's not use miniconda, since we don't want to be using anaconda products, let's rely on miniforge.

source/hpc.md
```diff
@@ -4,129 +4,311 @@ review_priority: "index"
...
+shared filesystem so compute nodes can access it and so the SLURM output file is
+written somewhere visible after the job completes.
+
+```bash
+#!/usr/bin/env bash
+#SBATCH --job-name=rapids-cudf
+#SBATCH --gres=gpu:1
+#SBATCH --time=01:00:00
+
+module load rapids/{{ rapids_version }}
```
ncclementi, line 329: Do you have to do the module loading in the script?

ncclementi (commented):
This is looking great, I left some comments to discuss and improve some things. 

I wonder if in a follow up we should try to get an example that uses multi-gpu, maybe with cudf polars or something.

source/hpc.md
jayavenkatesh19: I added this line back but removed the reference to InfiniBand as we are currently not using this in this page. 

Once we expand this to include infiniband/multi-GPU systems as a follow up, we can add this back.

source/hpc.md
jayavenkatesh19: I made this a more neutral message instead of just saying `conda` is the sole recommendation.

source/hpc.md
jayavenkatesh19: Added a section to install miniforge

source/hpc.md
jayavenkatesh19: Changed all the references from `miniconda` to `miniforge`

source/hpc.md
jayavenkatesh19: Yep, SLURM batch jobs load in a new shell on the compute node and modules loaded on the head node did not carry over in my testing. 

This also keeps it self-container and anyone can just run `sbatch` on this without worrying about modules

ncclementi (approved):
LGTM, just a heads up I did not test it directly. But everything seems reasonable. We can iterate over if we need to in the future

PR #677, opened 2026-04-15
Mention cuda.core.system as an option for getting NVML information.

mdboom:
Added reference to cuda.core.system for obtaining MIG instance names.


ncclementi (approved):
LGTM, thank you @mdboom

PR #695, opened 2026-06-01
Add conda_prefix to sagemaker script

### LINKED ISSUE #694: Broken sagemaker install needs conda_prefix
This is the same issue than azureml, without providing `CONDA_PREFIX` the env is not active in the notebook, and the installation is broken. 

This should be fixed by adding `--env CONDA_PREFIX "$CONDA_PREFIX"` to startup script.

ncclementi:
Closes https://github.com/rapidsai/deployment/issues/694


ncclementi:
I tested this and it works, @jayavenkatesh19  can I get a review so we can merge it.

jayavenkatesh19 (approved):

PR #698, opened 2026-06-03
Update GKE instructions to use NVIDIA GPU Operator

jayavenkatesh19:
This PR updates the GKE deployment page to use NVIDIA’s GPU Operator flow instead of relying on GKE’s default GPU driver and device plugins. 

While testing this page, I noticed that although the `cuda-vectoradd` test pod is working, the RAPIDS notebook failed to run with a missing CTK shared libraries error. Upon further investigation, I found that when using the GKE installed driver, services using the CTK required the `LD_LIBRARY_PATH` variable to be set ([From the GKE Docs](https://docs.cloud.google.com/kubernetes-engine/docs/how-to/gpus#cuda)). 

Using the NVIDIA GPU Operator solves this problem by configuring this out of the box. Instructions to use this on GKE are [linked here](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/google-gke.html#)

I tested the new drivers installed through the GPU Operator and the RAPIDS container now works as intended


source/cloud/gcp/gke.md
```diff
@@ -48,23 +49,88 @@ $ gcloud components install gke-gcloud-auth-plugin
 
 ```bash
 $ gcloud container clusters get-credentials rapids-gpu-kubeflow \
-    --region=us-central1-c
+    --zone us-central1-c
```
ncclementi, line 52: I guess using zone is fine here, we were already being very spcific when passing region. 

I do not recall if we have to be that specific, but I guess for this example is fine

source/cloud/gcp/gke.md
```diff
@@ -48,23 +49,88 @@ $ gcloud components install gke-gcloud-auth-plugin
...
+
+```bash
+kubectl apply -n gpu-operator -f - << EOF
+apiVersion: v1
+kind: ResourceQuota
+metadata:
+  name: gpu-operator-quota
+spec:
+  hard:
+    pods: 100
```
ncclementi, line 75: Do we need 100 pods? IIRC  is this a max allowed but we won't be creating that many right?

source/cloud/gcp/gke.md
```diff
@@ -48,23 +49,88 @@ $ gcloud components install gke-gcloud-auth-plugin
...
+
+Install the NVIDIA GPU Operator with the GKE-specific driver and toolkit paths.
+
+```bash
+$ helm repo add nvidia https://helm.ngc.nvidia.com/nvidia
+$ helm repo update
+$ helm install --wait gpu-operator \
+  -n gpu-operator \
+  nvidia/gpu-operator \
+  --version=v26.3.2 \
```
ncclementi, line 101: Is this a version that we will need to update every now and then? I think for now could be fine, but maybe if there is a way of using the latest always that would help

source/cloud/gcp/gke.md
```diff
@@ -48,23 +49,88 @@ $ gcloud components install gke-gcloud-auth-plugin
...
+```
+
+Verify that GPUs are allocatable on the node.
 
 ```console
-$ kubectl get po -A --watch | grep nvidia
-kube-system          nvidia-gpu-device-plugin-medium-cos-h5kkz                       2/2     Running   0          3m42s
-kube-system          nvidia-gpu-device-plugin-medium-cos-pw89w                       2/2     Running   0          3m42s
-kube-system          nvidia-gpu-device-plugin-medium-cos-wdnm9                       2/2     Running   0          3m42s
+$ kubectl get nodes -o=custom-columns='NAME:.metadata.name,GPU:.status.allocatable.nvidia\.com/gpu'
```
ncclementi, line 128: Before we were watching the pods, should we also do this here too?

I think we should check which pods are running, not just how many GPU each node has allocated. Specially because in the lines below we say "Once the pods are running"

source/cloud/gcp/gke.md
```diff
@@ -48,23 +49,88 @@ $ gcloud components install gke-gcloud-auth-plugin
...
+NAME                                                  GPU
+gke-rapids-gpu-kubeflow-default-pool-00000000-0000   2
 ```
 
-After GPU device plugin pods are in running state, you are ready to test your cluster.
+Once the GPU Operator Pods are running and GPUs are allocatable, you are ready to test your cluster.
 
 ```{include} ../../_includes/check-gpu-pod-works.md
```
ncclementi, line 135: Did you check this example runs, I noticed that it uses an image with. cuda 11.6 

` image: "nvidia/samples:vectoradd-cuda11.6.0-ubuntu18.04"`

do we know if this works? Maybe we should fine a way of updating this, if it currently runs, we can leave it and open a separate ticket.

source/cloud/gcp/gke.md
jayavenkatesh19: I am using zone here because the cluster creation command uses `zone`

source/cloud/gcp/gke.md
jayavenkatesh19: This is the recommended number according to the docs here: https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/google-gke.html

Like you said, its the max allowed so doesn't hurt to do this in my opinion

source/cloud/gcp/gke.md
jayavenkatesh19: While theoretically using omitting version here should give us the latest, but the instructions in the docs page for the GPU Operator pin this specific version: https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/google-gke.html#using-the-google-driver-installer

I'm inclined to stick with this because we tested this out and we know this works.

source/cloud/gcp/gke.md
jayavenkatesh19: We are watching the pods here as well (the previous command), but even though a few of the GPU operator pods are running, the RAPIDS container was not able to get a GPU allocation until the node details showed that GPUs were available.

Think this is an additional verification that is good to have

source/cloud/gcp/gke.md
jayavenkatesh19: This works currently, but we should open a separate ticket for updating this globally.

ncclementi: @jayavenkatesh19 would you mind opening a ticket for 26.08

ncclementi (approved):
LGTM! Thanks you for the work

Issue #710, opened 2026-06-23
Update Vertex AI guide to remove note con cuda toolkit 11


ncclementi:
We have a this note: 

<img width="775" height="364" alt="Image" src="https://github.com/user-attachments/assets/67b5ef4c-cce2-48ae-a761-334fc28a16ce" />

I believe this is no longer the case, that being said, the default env is broken at the moment, see #709 . Once that is fixed, make sure to test the installation instructions via pip and update accordingly.

Issue #709, opened 2026-06-23
Vertex AI default environment has broken cupy installation


ncclementi:
In the altest Vertex AI env, the cuda toolkit version is 13

```bash
ls -ld /usr/local/cuda*
```

```bash
lrwxrwxrwx  1 root root   21 Jun  1 04:54 /usr/local/cuda -> /usr/local/cuda-13.0/
drwxr-xr-x 17 root root 4096 Jun  1 04:55 /usr/local/cuda-13.0
```

but when checking pip list they have a cuda 12 installation
```bash
cupy-cuda12x                          13.6.0
```
Which is a broken env:

```python 
import cupy as cp
X = cp.array([[-1, -1], [-2, -1], [-3, -2], [1, 1], [2, 1],[3, 2]], cp.float32)
X**2
```

```
RuntimeError: CuPy failed to load libnvrtc.so.12: OSError: libnvrtc.so.12: cannot open shared object file: No such file or directory
```

ncclementi:
This was closed by: https://github.com/rapidsai/deployment/pull/725

Issue #520, opened 2025-02-11
[AWS] RAPIDS 25.02 conda packages cannot be installed on SageMaker Notebook Instances: requires GLIBC 2.28+


jameslamb:
## Description

As of this writing, Amazon SageMaker Notebook Instances only support Amazon Linux 2: https://docs.aws.amazon.com/sagemaker/latest/dg/nbi-al2.html

That distribution uses GLIBC 2.26.

As a result of https://github.com/rapidsai/build-planning/issues/131, RAPIDS 25.02+ packages required GLIBC 2.28+.

Taken together, this means that **RAPIDS 25.02 conda packages cannot be installed on any of the "platforms" supported by Sagemaker Notebook Instances**.

It seems like on an `ml.p3.2xlarge` (the instance type we recommend on https://docs.rapids.ai/deployment/nightly/cloud/aws/sagemaker/), using whatever the corresponding AMI is for `Amazon Linux 2, Jupyterlab 4`, this fails.

## Reproducible Example

I created a SageMaker Notebook Instance in `us-west-2`, following the docs at https://docs.rapids.ai/deployment/nightly/cloud/aws/sagemaker/.

* instance type: `ml.p3.2xlarge`
* "platform": `Amazon Linux 2, Jupyterlab 4`

Put the following into a lifecycle configuration

```shell
mamba create -y -n rapids -c rapidsai-nightly -c conda-forge -c nvidia rapids=25.02 python=3.12 cuda-version=12.5 \
    boto3 \
    ipykernel \
    'sagemaker-python-sdk>=2.239.0'
```

That failed, like this:

```text
Looking for: ['rapids=25.02', 'python=3.12', 'cuda-version=12.5', 'boto3', 'ipykernel', "sagemaker-python-sdk[version='>=2.239.0']"]

conda-forge/linux-64                                        Using cache
conda-forge/noarch                                          Using cache
nvidia/linux-64                                             Using cache
nvidia/noarch                                               Using cache
pytorch/linux-64                                            Using cache
pytorch/noarch                                              Using cache
rapidsai-nightly/linux-64                                     No change
rapidsai-nightly/noarch                                       No change
Could not solve for environment specs
The following package could not be installed
└─ rapids 25.02**  is not installable because it requires
   └─ cuvs 25.02.* , which requires
      └─ __glibc >=2.28,<3.0.a0 , which is missing on the system.
```

<details><summary>output of 'conda info' (click me)</summary>

```text
[...]

PR #719, opened 2026-07-24
Default the deployment docs to CUDA 13

### LINKED ISSUE #612: Update to CUDA 13
## Description

RAPIDS 25.10 will be the first release to include CUDA 13 support: https://github.com/rapidsai/build-planning/issues/208

Wherever possible, docs here should point users towards CUDA 13 packages and container images.

## Benefits of this work

* all the benefits of CUDA 13 relative to previous versions: https://developer.nvidia.com/blog/whats-new-and-important-in-cuda-toolkit-13-0/
* may help us discover deployment targets that can't / don't yet support CUDA 13, starting the process of helping those targets update sooner

## Acceptance Criteria

* CUDA 13 images and the packages are the default for all docs where that's possible today

## Approach

Update default image tags to `cuda13.0` here:

https://github.com/rapidsai/deployment/blob/41b835e15179db0c504d41e0175919f0c0cae433/source/conf.py#L27

Look through the rest of the docs for other recommendations about container images, AMIs, etc. for opportunities to update to CUDA 13.

## Notes

Related CUDA version issues:

* #584
* #583

jayavenkatesh19:
Closes #612, #709 and #710

RAPIDS and most of our CSPs have moved to CUDA 13, while our deployment docs still default to CUDA 12. This PR makes CUDA 13 the default recommendation. 

While doing this, I pulled in CUDA and Python version strings into `conf.py` so that future version bumps can be small edits limited to this file instead of having to grep through all of the docs to replace the relevant container tags/conda/pip install instructions. 

Almost all the platforms we have docs for are on CUDA 13, so changes are pretty straightforward except for a couple of caveats:

- AzureML stays on CUDA 12.2. They are still using Driver version 535.xx which is too old to support CUDA 13. I've added a note explaining this on the page
- Databricks moved to CUDA 13, but they still ship a CuPy-cu12 in their runtime, so the new init script uninstalls that before installing CUDA 13 wheels (need to test this)
- Sagemaker's previous version pin can be dropped, as #520 has been fixed. 

All changes are limited to cloud/platforms and the HPC docs. Workflow examples have been left untouched because they could break on a CUDA version bump and would be hard to test out.


jameslamb (approved):
Nice, I support it!

source/conf.py
```diff
@@ -24,35 +24,46 @@
...
-        "rapids_conda_packages": f"rapids={stable_version} python=3.13 'cuda-version>=12.0,<=12.9'",
+        "rapids_conda_packages": f"rapids={stable_version} python={python_version} '{stable_cuda_range}'",
         "rapids_pip_index": "https://pypi.nvidia.com",
         "rapids_pip_version": stable_version,
-        # SageMaker Notebook Instance examples need to stay pinned to an older RAPIDS until this is resolved:
-        # https://github.com/rapidsai/deployment/issues/520
-        "rapids_sagemaker_conda_packages": f"rapids={stable_version} python=3.12 cuda-version=13",
+        "rapids_cuda_major": cuda_major,
+        "rapids_cuda_version_range": stable_cuda_range,
+        # AzureML is pinned to CUDA 12 currently (driver 535.x supports only CUDA 12.x).
```
ncclementi, line 50: I'm curious if we should support this, because 535 is EOL already. Or should we just put a big warning, mentioning the issue in AzureML. @jameslamb any advice here?

source/cloud/azure/azureml.md
```diff
@@ -45,6 +45,11 @@ from the script, the VM will boot and inspect it by running each line of the scr
 to see where it fails.
 ```
 
+```{note}
```
ncclementi: this should be a warning, and mentioned that 535 is EOL, and we can't guarantee proper functionality. 
@jameslamb is that too agresive?

source/cloud/gcp/vertex-ai.md
```diff
@@ -25,8 +25,7 @@ Once the instance has started select **OPEN JUPYTER LAB** and at the top of a no
 ```{warning}
 Installing RAPIDS via `pip` in the default environment is [not currently possible](https://github.com/rapidsai/deployment/issues/517), for now you must create a new `conda` environment.
 
-Vertex AI currently ships with CUDA Toolkit 11 system packages as of the [Jan 2025 Vertex AI release](https://cloud.google.com/vertex-ai/docs/release-notes#January_31_2025).
-The default Python environment also contains the `cupy-cuda12x` package. This means it's not possible to install RAPIDS package like `cudf` via `pip` as `cudf-cu12` will conflict with the CUDA Toolkit version but `cudf-cu11` will conflict with the `cupy` version.
+Vertex AI now ships with CUDA 13 system packages, and the default Python environment contains a `cupy-cuda12x` package. Installing RAPIDS via `pip` into that environment conflicts with the preinstalled `cupy`, so create a dedicated `conda` environment (below) instead.
```
ncclementi, line 28: Does this mean that this is still broken? I though they fix it. They sent an email saying this was fixed. Or we haven't tested it and we are relying on the previous release experience.

source/platforms/databricks.md
```diff
@@ -26,10 +26,15 @@ Navigate to the top-left **Workspace** tab and click on your **Home** directory
 #!/bin/bash
 set -e
 
+# The Databricks ML runtime preinstalls cupy-cuda12x. RAPIDS CUDA 13 wheels
+# need the matching cupy-cuda13x build, so replace it before installing.
+pip uninstall -y cupy-cuda12x
+
```
ncclementi, line 32: We should ping them to start using cuda 13 packages. @jrbourbeau Woudl you be able to help with this? 

They already have drivers that support cuda 13 but are still installing cuda 12 packages, can we ask them to make the bump?

source/cloud/gcp/vertex-ai.md
jayavenkatesh19: This is based on information from our environment tracker table. This is something we do need to test before release and we can amend it to what we find.

source/cloud/azure/azureml.md
jayavenkatesh19: Made this into a warning, and I think when EOL is reached, it is important to convey that to users.

ncclementi (approved):
LGTM, we can merge and start release testing, we will fix as we test.

source/cloud/azure/azureml.md
jameslamb: Sorry for taking too long to respond here. I'm glad it ended up with a warning. I would probably:

* link to the evidence for that driver being EOL (https://docs.nvidia.com/datacenter/tesla/drivers/supported-drivers-and-cuda-toolkit-versions.html)
* remove the "we cannot guarantee...." language

RAPIDS is, as of this writing, still getting tested on driver v535:

* (private link, sorry): https://github.com/nv-gha-runners/enterprise-runner-configuration/blob/main/docs/runner-groups.md#nv-gpu-arm64-a100-1gpu
* see `nvidia-smi` output here for example: https://github.com/NVIDIA/cuvs/actions/runs/30433470230/job/90617582436

```text
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.309.01             Driver Version: 535.309.01   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  Tesla V100-PCIE-32GB           On  | 00000000:0A:00.0 Off |                    0 |
| N/A   40C    P0              24W / 250W |      0MiB / 32768MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
```

And we still list that version in our platform support docs: https://docs.rapids.ai/platform-support/

image

source/conf.py
jameslamb: I think it's worth still supporting this, for as long as RAPIDS has new packages supporting that driver version.

PR #676, opened 2026-04-15
Fraud detection workflow example

jayavenkatesh19:
Towards #667 

Adds a new workflow example demonstrating how to wrap the [NVIDIA Financial Fraud Detection AI
Blueprint](https://github.com/NVIDIA-AI-Blueprints/financial-fraud-detection) in production infrastructure
using Prefect (orchestration), MLflow (experiment tracking + model registry), and Triton (champion/challenger
 serving with native versioning).

The notebook walks through the architecture, each pipeline stage, running and monitoring the pipeline, and
scaling to multiple machines.

PR #721, opened 2026-07-28
Fixes after testing on Databricks

jayavenkatesh19:
Merging #719 with additional templating created an issue with nightly version string. This PR fixes that. 

Also tested out all the instructions on Databricks 18 LTS runtime with Machine Learning enabled, and `cupy` is not installed on this environment, which makes the `init-script` shorter. 

Going through instances of nightly pip versioning, I noticed that modal was not using the pip versioning while installing with `uv`, so this PR updates that

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -59,10 +59,6 @@ versions = {
         "rapids_conda_channels": "-c rapidsai-nightly -c conda-forge",
         "rapids_conda_packages": f"rapids={nightly_version} python={python_version} '{nightly_cuda_range}'",
         "rapids_pip_index": "https://pypi.anaconda.org/rapidsai-wheels-nightly/simple",
-        # Callers pin with '==', so this must stay a bare version. The trailing
-        # '>=0.0.0a0' is an unbounded no-op whose only job is to put a prerelease
-        # marker in the specifier set, which is what makes pip consider nightly
-        # prereleases without needing '--pre'.
         "rapids_pip_version": f"{nightly_version}.*,>=0.0.0a0",
         "rapids_cuda_major": cuda_major,
         "rapids_cuda_version_range": nightly_cuda_range,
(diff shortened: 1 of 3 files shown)


source/conf.py
```diff
@@ -59,7 +59,7 @@
         "rapids_conda_channels": "-c rapidsai-nightly -c conda-forge",
         "rapids_conda_packages": f"rapids={nightly_version} python={python_version} '{nightly_cuda_range}'",
         "rapids_pip_index": "https://pypi.anaconda.org/rapidsai-wheels-nightly/simple",
-        "rapids_pip_version": f"~={nightly_version}.0a0",
+        "rapids_pip_version": f"{nightly_version}.*,>=0.0.0a0",
```
ncclementi, line 62: In the modal page, this renders as 

```bash
image = ctk_image.uv_pip_install(
    "cudf-cu13==26.08.*,>=0.0.0a0",
).env({"CUDF_PANDAS_RMM_MODE": "async"})
```

But I think this is incorrect, because pip nightlies, 
1. need the extra index 
2. when looking at the install rapids selector, the limits for the nightly are marked as `"cudf-cu13>=26.8.0a0,<=26.8"`

ncclementi (commented):
Not sure if we can fix this with templating, or need to modify something to include the extra index somehow. when pip is used.

source/conf.py
jayavenkatesh19: I added the index in the install command on Modal. 

Also I tested this string out, and looked at historical usage of this. Before I switched this over in #676, this was how we used to install nightlies and I resolved both locally to find that they pointed to the same wheel. 

The notation changed in that PR is probably easier to read, but does not work with templating changes now, so we're switching back. Functionally, everything should remain the same.

ncclementi: Noting that, we only need the index for the nigthly, because we are only installing cudf,  but I guess this is a way to make things work as a generalized piece of code.

ncclementi (approved):
Approved, we one comment. 

For the modal example, we only need the index for the nigthly, because we are only installing cudf, but I guess having the index explicit, makes easier the generalization of the  code.

PR #724, opened 2026-07-29
X-ORG-722: Publish API docs to docs.nvidia.com

### LINKED ISSUE #722: [X-ORG] Publish docs to docs.nvidia.com via new shared workflow
As part of publishing API docs to docs.nvidia.com, we need to integrate a [new shared action](https://github.com/rapidsai/shared-actions/tree/main/publish-docs) into the build workflow.

This integration will take the docs publish by the current docs-building job and push them to the docs.nvidia.com S3 bucket. It does not automatically make those docs go live - that needs to be handled out-of-band by the docs team themselves.

XREF: https://github.com/rapidsai/build-infra/issues/380

josephine-wolf-oberholtzer:
Closes #722 

Let's publish these docs to https://docs.nvidia.com/datascience/deployment.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -49,8 +49,10 @@ jobs:
         run: echo "VERSION=${{ startsWith(github.event.ref, 'refs/tags/') && 'stable' || 'nightly' }}" >> "$GITHUB_OUTPUT"
 
       - name: Sync HTML files to S3
+        env:
+          VERSION: ${{ steps.set-version.outputs.VERSION }}
         if: ${{ github.repository == 'rapidsai/deployment' && (github.event_name == 'push' || github.event_name == 'workflow_dispatch') }}
-        run: aws s3 sync --no-progress --delete build/dirhtml "s3://rapidsai-docs/deployment/html/${{ steps.set-version.outputs.VERSION }}"
+        run: aws s3 sync --no-progress --delete build/dirhtml "s3://rapidsai-docs/deployment/html/${VERSION}"
 
       - name: Publish to docs.nvidia.com
         if: ${{ github.repository == 'rapidsai/deployment' && (github.event_name == 'push' || github.event_name == 'workflow_dispatch') }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-deploy.yml
```diff
@@ -44,8 +44,31 @@ jobs:
...
+          akamai-host: ${{ secrets.NVIDIA_DOCS_AKAMAI_HOST }}
+          akamai-request-name: rapidsai-deployment-${{ steps.set-version.outputs.VERSION }}
+          dry-run: false
+          source-path: build/dirhtml
+          target-aws-access-key-id: ${{ secrets.NVIDIA_DOCS_AWS_ACCESS_KEY_ID }}
+          target-aws-region: ${{ secrets.NVIDIA_DOCS_AWS_REGION }}
+          target-aws-secret-access-key: ${{ secrets.NVIDIA_DOCS_AWS_SECRET_ACCESS_KEY }}
+          target-s3-bucket: ${{ secrets.NVIDIA_DOCS_S3_BUCKET }}
+          target-s3-key-suffix: ${{ steps.set-version.outputs.VERSION }} # stable or nightly
+          target-s3-key: datascience/deployment
```
ncclementi, line 74: I'm not sure we necessarily fall under datascience. Does this target s3 key, maps to what the final url would be? 
If yes, @josephine-wolf-oberholtzer do we get to choose this, or not at all? 

@mmccarty Do you have any input here?

ncclementi (commented):
This looks pretty good, before merging, We have a question regarding the where the final url would land. 

The deployment docs, go beyond the scope of datascience, it would be great to know if this is modifiable or not. See comment in line

.github/workflows/build-and-deploy.yml
taureandyernv: We did discuss that the deployment docs should go under `/datascience/deploy`.  While it is modifiable, where do you propose it should be?

josephine-wolf-oberholtzer: 1) Yes, the S3 key maps to the final URL AFAIUI.
2) `datascience` was the suggested naming convention from Alex. Let's take this to the migration project channel for hashing out? #swrapids-nvidia-migration.

ncclementi:
After discussion, it seem like `/datascience` is the placement for now. I'll approve this and we will revisit in the future if needed.

ncclementi (approved):

josephine-wolf-oberholtzer:
I'll have to come back to this - need to be able to run it in a CI container for the action to work.

PR #728, opened 2026-07-31
X-ORG-722: Ensure gha-tools are installed

josephine-wolf-oberholtzer:
Contributes to #722 

Noticed a couple problems in the [previous PR](https://github.com/rapidsai/deployment/pull/724)'s build post-merge: https://github.com/rapidsai/deployment/actions/runs/30579742034

- The `publish-docs` action assumes we're inside a CI container because it doesn't use `sudo` for its `apt-get install`. This is now fixed via https://github.com/rapidsai/shared-actions/pull/133
- We also need https://github.com/rapidsai/gha-tools installed.

I tried to run the entire workflow in a CI container, but those have Python 3.14, while this build appears to expect 3.12, and there are some Python package installation problems I don't have bandwidth for at the moment.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -21,6 +21,7 @@ jobs:
   build:
     name: Build (and deploy)
     runs-on: ubuntu-latest
+
     steps:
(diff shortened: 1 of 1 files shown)


ncclementi (approved):
LGTM

PR #733, opened 2026-08-04
added new Brev launchable instructions

jayavenkatesh19:
Updates the Brev launchable instructions to reflect the new UI changes on Brev. 

While testing this out, I also found a few bugs and made some changes in the old Docker Compose YAML file:

- `backend` (now renamed to `jupyter`) was not nested under services and was failing YAML parsing
- `--ip=127.0.0.1` changed to `--ip=0.0.0.0`. I could not reach the earlier Jupyter endpoint over the internet and this configuration change worked
- Dropped working directory and `--notebook-dir=/notebooks`. This was showing up in Jupyterlab as an empty folder and not with the examples as intended
- Mounting any code repositories as a volume and using Brev's source section to clone it. This ensures that the files in the repo are visible on Jupyterlab
- `EXTRA_CONDA_PACKAGES` is now a launch configuration variable with the value being passed to the container. This way users can customize their packages at deploy time rather than the Launchable creation time

I added screenshots of every single section, and tested everything out end to end. This new configuration ensures a healthy Jupyterlab, with access to any cloned repositories and all of the example notebooks.


source/cloud/nvidia/brev.md
```diff
@@ -48,65 +48,112 @@ our [Single-cell Analysis Blueprint](https://github.com/NVIDIA-AI-Blueprints/sin
...
+    environment:
+      - EXTRA_CONDA_PACKAGES # Value comes from a launch parameter of the same name
+    ports:
+      - "8888:8888" # Expose JupyterLab
+    volumes:
+      # Repo cloned by the Source section, mounted alongside the example notebooks.
+      # Remove this entry if you are not adding a repository.
+      - /home/ubuntu/<name_of_your_github_repo>:/home/rapids/notebooks/<name_of_your_github_repo>
+    user: root
+    command: jupyter-lab --notebook-dir=/home/rapids/notebooks --ip=0.0.0.0 --no-browser --allow-root --NotebookApp.token='' --NotebookApp.allow_origin='*'
```
ncclementi, line 99: I understand the 0.0.0.0 but I'm not sure if that's secure enough. Should we ask the  Brev team about this?

source/cloud/nvidia/brev.md
```diff
@@ -48,65 +48,112 @@ our [Single-cell Analysis Blueprint](https://github.com/NVIDIA-AI-Blueprints/sin
...
+    shm_size: 1g
+    deploy:
+      resources:
+        reservations:
+          devices:
+            - driver: nvidia
+              count: all
+              capabilities: [gpu]
+    environment:
+      - EXTRA_CONDA_PACKAGES # Value comes from a launch parameter of the same name
```
ncclementi, line 91: If we give put thsi as an example, don't we need `:` and some packages to show teh exmaple? 

I'm worried this dockerfile won't work as is.

source/cloud/nvidia/brev.md
```diff
@@ -48,65 +48,112 @@ our [Single-cell Analysis Blueprint](https://github.com/NVIDIA-AI-Blueprints/sin
...
+    ports:
+      - "8888:8888" # Expose JupyterLab
+    volumes:
+      # Repo cloned by the Source section, mounted alongside the example notebooks.
+      # Remove this entry if you are not adding a repository.
+      - /home/ubuntu/<name_of_your_github_repo>:/home/rapids/notebooks/<name_of_your_github_repo>
```
ncclementi, line 97: I assume we remove this to mount the whole repo and not map it to the notebooks directory right? Just wanted to make sure

source/cloud/nvidia/brev.md
```diff
@@ -48,65 +48,112 @@ our [Single-cell Analysis Blueprint](https://github.com/NVIDIA-AI-Blueprints/sin
...
+
+#### Launch Parameters
+
+This is an optional step. Launch parameters collect values when someone deploys your Launchable and expose them as environment variables during container startup.
+
+This is useful for adding your own libraries to the container via `EXTRA_CONDA_PACKAGES`. Add a launch parameter of that name, set any defaults as you see fit, and then list the variable by name in the compose file so its value is passed into the container. While deploying your Launchable, you can then install extra packages each time without editing the Launchable itself:
+
+```yaml
+environment:
+  - EXTRA_CONDA_PACKAGES
```
ncclementi, line 128: Have you checked once the instance is launched if the extra packages installed work? 
To see if they were properly installed?

ncclementi (commented):
Looks pretty good. I think the docker compose should be self contained. like the equivalent to launching things via UI.

source/cloud/nvidia/brev.md
jayavenkatesh19: We can ask them about this yes, but while debugging I couldn't find a way to make the `127.0.0.1` work. I even tried without the `--NotebookApp.token=' '` which disables authentication. We can ask the Brev team about how internet access/firewall rules are setup

source/cloud/nvidia/brev.md
jayavenkatesh19: We don't need it, as this variable is populated from the value from the Launch Parameters section. I added hdbscan and umap in launch parameters and I could import those packages in the Notebooks

source/cloud/nvidia/brev.md
jayavenkatesh19: The reason for doing this was to keep the example notebooks visible to the user. If we map to the notebooks directory, users can see the `cudf`, `cugraph`, `cuml` examples folders along with the cloned repo added as the fourth folder in the root of the mounted path

source/cloud/nvidia/brev.md
jayavenkatesh19: Yep I could import them into the notebook without any errors.

ncclementi (approved):
pending understanding on the 0.0.0.0 change.

jayavenkatesh19:
Confirmed with the Brev team that they switched from using Cloudflare for secure links to an internal tool called Skybridge which only supports external access through the `0.0.0.0` IP address. 
Merging this PR!

PR #742, opened 2026-08-13
improvements to the LULC workflow example

### LINKED ISSUE #641: Enhancements for the LULC classification workflows example
The follow elements have to be added to the notebook to make it more non-GIS user friendly. 

- [ ] Environment creation and requirements specification
- [ ] Further explanation about CRS and difference between using EPSG codes like 4326 etc
- [ ] A markdown cell describing how to create your own AOI on an online GeoJSON tool such as [geojson.io](https://geojson.io/#map=2/0/20)
- [ ] A markdown cell describing hardware details and timing charts (can reuse the same ones from the PyData Boston talk)

jayavenkatesh19:
Closes #641


source/examples/lulc-classification-gpu/environment.yml
```diff
@@ -0,0 +1,30 @@
+name: rapids-lulc
+channels:
+  - rapidsai
+  - conda-forge
+  - nvidia
+dependencies:
+  - python=3.13
+  - cuda-version>=13.0,<=13.2
+  - rapids=26.08
```
ncclementi: Do we want to install all rapids? since we are already installing bunch of other stuff.  Do we need all rapids?

ncclementi (approved):
@jayavenkatesh19 and I did a final path review together, this is good to go.

PR #753, opened 2026-09-08
Update links to docs for libraries from rapidsai -> nvidia

### LINKED ISSUE #745: Replace `rapidsai/<library>` references with `NVIDIA/<library>`
As we migrate to nvidia docs. We need to update github repository references from `rapidsai/<library>` to `NVIDIA/<library>`/ . 

As well as references to documentation pages pointing to rapids docs, instead of nvidia docs. 

There will be some exceptions like GHCR image paths use the required lowercase nvidia/<library> namespace.

ncclementi:
Closes https://github.com/rapidsai/deployment/issues/745 

I used this AI generated script, to do some replacements and track cases where we don't have a solution yet. 

https://gist.github.com/ncclementi/654952edd9d43ae796b2f748d018af26

`./scripts/update_docs_links.sh` to see possible replacements links and ones with no case 

`./scripts/update_docs_links.sh --write` to do replacements

Note: I think most links are correct, we should implement lychee to capture if any ended up broken, but I'm confident in most replacements


jayavenkatesh19 (approved):

source/guides/azure/infiniband.md
```diff
@@ -255,7 +255,7 @@ Accept the default and allow conda init to run.
 
 Then start a new shell.
 
-Create a conda environment (see [UCXX](https://docs.rapids.ai/api/ucxx/nightly/install/) docs)
+Create a conda environment (see [UCXX](https://docs.nvidia.com/ucxx/latest/install/) docs)
```
ncclementi: ucxx is not being migrated, so leave as is for now

```suggestion
Create a conda environment (see [UCXX](https://docs.rapids.ai/api/ucxx/nightly/install/) docs)
```

PR #754, opened 2026-09-08
Publish versioned docs to docs.nvidia.com with a version switcher

jayavenkatesh19:
Addition to #722 

This PR is intended to move publishing docs away from `stable`/`nightly` versioning and onto the `latest`/`YY.MM` versioning scheme. 

- Pushes to main publish to `datascience/deployment/latest`
- vYY.MM.PP tags publish to `datascience/deployment/YY.MM` (alpha tags do not publish)
- A `version.json` is published next to those directories which enables the navbar version switcher. 

Both uploads go through the `publish-docs` and `publish-one-doc` shared actions. The workflow itself has no logic in it, the Sphinx build writes a small ephemeral manifest at `build/publish/publish.env` which specifies where the files have to be published to which is picked up by the workflow. 

The `version.json` is created from a new Sphinx extension called `rapids_docs_publishing`. It runs only when the workflow sets `RAPIDS_DOCS_PUBLISH=true`, so local and ReadTheDocs builds do have show any switcher on the builds. 

This PR also stops publishing to the old `rapidsai-docs` bucket and we will setup redirects on `rapidsai/docs` to redirect all links from `docs.rapids.ai/deployment/*` to `docs.nvidia.com/datascience/deployment/*`

Can only be merged after https://github.com/rapidsai/shared-actions/pull/142 is merged.  [EDIT: This has been merged]


.github/workflows/build-and-deploy.yml
```diff
@@ -33,48 +49,52 @@ jobs:
...
-        if: ${{ github.repository == 'rapidsai/deployment' && (github.event_name == 'push' || github.event_name == 'workflow_dispatch') }}
-        with:
-          role-to-assume: ${{ vars.AWS_ROLE_ARN }}
-          aws-region: ${{ vars.AWS_REGION }}
-          role-duration-seconds: 3600 # 1h
-
-      - name: Set version
-        id: set-version
-        run: echo "VERSION=${{ startsWith(github.event.ref, 'refs/tags/') && 'stable' || 'nightly' }}" >> "$GITHUB_OUTPUT"
+        run: RAPIDS_DOCS_RELEASE_TAGS="$(git tag --list)" uv run make dirhtml SPHINXOPTS="-W --keep-going -n"
```
ncclementi, line 54: Just checking did we end up pushing to the rapids S3? are we ready to stop publishing the rapids docs?

ncclementi:
Just to check, in the current read the docs generated in CI we do have in the nav bar the `dev` tag. I'm not sure if we want to keep that or if we don't care. 

Screenshot 2026-09-09 at 11 31 59 AM

.github/workflows/build-and-deploy.yml
bdice: As soon as we get something on docs.nvidia.com we can stop publishing to docs.rapids.ai and add redirects.

bdice (approved):
Approving with very light review since this has to be merged to be tested.

jayavenkatesh19:
Answering @ncclementi's question, we will now not have a switcher on local/ReadTheDocs builds. 

We previously had a switcher which was defined in a HTML file (see the older approach in PR changes), but the new NVIDIA theme has the option to enable the switcher only when a `versions.json` file is present in the build directory. We create that file in the new Sphinx extension added by the PR, and push it to the S3 bucket so that the `docs.nvidia.com` site does have the switcher.

ncclementi (approved):
Let's merge this and see what happens, as we'll need to iterate

